PR #127, opened 2012-10-17
ADDED - `scale` parameter to options

glesperance:
Hey there, 

I needed to make an image out of an element in my webapp and had to output an image that had a scaling factor applied to it.

I couldn't get any success with the `width` and `heigh`t properties so I added a `scale` parameter that can be passed as an option.

Scale can either be a number, or an object of the form `{ x : scaleX, y : scaleY }`. The resulting image should also be of good quality as the scaling factor is applied before the drawing.

Cheers,

Gabriel


stefanello:
@niklasvh 
any news for adding support for css transform:scale() ?
Thx!

haoqi0818:
in the most cases, the body's style is "overflow:hidden", so just set "transform:scale(2)" to aim element is not work as expected

PR #180, opened 2013-03-24
Upgraded to Grunt 0.4.0 and added Safari for Mac bug fix.

djfarrelly:
Upgraded the `Gruntfile.js` to be compatible with Grunt 0.4.x and added appropriate dependencies to `package.json`.  

Added a try/catch to fix Safari's issue with writing read only properties to pseudo elements.

Altered the `selenium.js`'s markdown function to skip report json files that do not exist instead of failing.  I've also added the lines to export a true markdown format file instead of html tables.  It works, but is currently commented out.

All qunit tests pass and I've added the new Mac benchmarks to the test readme.md.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -50,8 +50,7 @@ Run the full build process (including lint, qunit and webdriver tests):
 
 Skip lint and tests and simply build from source:
 
-    $ grunt concat
-    $ grunt min
+    $ grunt build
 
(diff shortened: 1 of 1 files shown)


luck2011:
why safari still cannot work if you select part of the DOM rather than document.body

PR #253, opened 2013-08-05
z-index overhaul. relative above static; negative z-index

arrix:
I reworked the z-index part.
1. when stacking without z-index, positioned > floated > normal flow
2. supports negative z-index
3. new stacking context formed when opacity < 1 (standard)
4. new stacking context formed for position:fixed (no standard yet, done in mobile webkit and chrome 22+)

Added 6 z-index tests. All z-index tests are passing in Chrome and Firefox in OSX. 
Please review~


jackuro:
hi guys i have a problem i am trying to mix and image using css mix-blend-mode: multiply;, but in the moment of the screensshot it only bring me the imagen without the mix-blend-mode: multiply; can you hel me please?

PR #261, opened 2013-08-15
Added multiline + text wrap support for textarea

### LINKED ISSUE #259: Textarea renders multiline incorrectly
The current renderFormValue and renderText methods don't take into account the fact that "\n" characters are rendered on textarea as new line. 

I'm hacking into the source code to fix this, so anyone has a good idea which component i can reuse or which direction should i go to make this happen?

thomasttvo:
resolve #259

### DIFF
--- a/src/Parse.js
+++ b/src/Parse.js
@@ -164,33 +164,98 @@ _html2canvas.Parse = function (images, options) {
     };
 
     if (_html2canvas.Util.trimText(textNode.nodeValue).length > 0) {
-      textNode.nodeValue = textTransform(textNode.nodeValue, getCSS(el, "textTransform"));
-      textAlign = textAlign.replace(["-webkit-auto"],["auto"]);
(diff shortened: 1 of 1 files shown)


moxet:
I am still facing the same problem with the latest build, how to fix multiple lines?

thomasttvo:
@moxet try the fix that I submitted in this pull request. You need to apply it manually though.

moxet:
Sorry to bother you but i have limited knowledge of solving it manually. I don't have the structure of parse.js and others. I have html2canvas.js file only. Where can i make the changes?

thomasttvo:
sorry @moxet, I've over looked the fact that this file was far out of date. The new html2canvas code is very different now, so my changes wouldn't help you with this.

charles-dexter-ward:
line-height with a value of normal is not taken into consideration. It should be calculated in some way based on the font. Not so simple because every browser has its own method to calculate the line-height when it is setted as "normal".

PR #320, opened 2014-01-10
Fixed IndexSizeError in IE when the size of the image exceedes 8192 pixels

oosterholt:
Patch for the IndexSizeError IN IE.


brenosilver:
The drawImage and putImageData is only working for <= 2048px on ie9. if I enable "use software rendering instead of GPU rendering" this value goes up to 8192px.
I can't depend on enabling this option for obvious reason. 

I need at least those 8192px, My company needs this asap for auditing reasons.

Does anyone have any idea why this is happening?

PR #499, opened 2015-01-13
add support for multiple Text Shadows

mudcube:
add support for multiple text-shadows, e.g.:

text-shadow: 0 0 5px #00f, 2px 2px 0 #f00;

add support for single text-shadow on text elements whose color is semi-transparent; Shadows must be rendered separately from text for Canvas to achieve this effect, otherwise everything ends up semi-transparent.


mudcube:
I don't quite get this Travis CI build failure. It says ECONNREFUSED on browser tunnel, but not sure how that has to do with the code failing to check properly (or perhaps this is a normal thing with Travis CI? I don't have experience there). Here's where the error occurs, ideas how to solve?

Running "webdriver" task
 STARTING (1/65) chrome-34-Windows 7 tests/cases/background/clip.html 
 STARTING (1/65) firefox-15-Windows 7 tests/cases/background/clip.html 
 STARTING (1/65) internet explorer-9-Windows 7 tests/cases/background/clip.html 
 ERROR chrome-34-Windows 7 [init({"browserName":"chrome","platform":"Windows 7","version":"34","tunnel-identifier":"236.1","name":"576eb6a2c7","build":"236"})] connect ECONNREFUSED
Done
 ERROR firefox-15-Windows 7 [init({"browserName":"firefox","version":"15","platform":"Windows 7","tunnel-identifier":"236.1","name":"576eb6a2c7","build":"236"})] connect ECONNREFUSED
 ERROR internet explorer-9-Windows 7 [init({"browserName":"internet explorer","version":"9","platform":"Windows 7","tunnel-identifier":"236.1","name":"576eb6a2c7","build":"236"})] connect ECONNREFUSED
Warning: Task "webdriver" failed. Use --force to continue.
Aborted due to warnings.
npm ERR! Test failed.  See above for more details.
npm ERR! not ok code 0

niklasvh:
Thanks for the PR!

The travis webdriver tests always fail for pull requests (http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests).

Few issues:

 - The whitespace should be 4 spaces instead of tab.
 - Whats the purpose of the `dx` variable and why is it needed? The current test doesn't work at all for me with that dx value added/subtracted from the different values.
 - As the text gets rendered multiple times and the fill color is hardcoded to black, won't this break rgba colors which aren't 100% opacity?
 - `NodeParser` shouldn't directly access the canvas context, instead use the `setVariable`  or other methods / properties in the `CanvasRenderer`. 

I've added some more tests to https://github.com/niklasvh/html2canvas/blob/master/tests/cases/text/shadow.html
(including some with the opacity issue) but would be nice to have some more.

mudcube:
2)  The dx value is needed as an offset so the shadow is separated from the text (it should push the text out of bounds, while the shadow stays in place). This is because in canvas context there is no way to create a shadow of a certain color without also creating text of the same color. While in CSS you can specify multiple text-shadows, so in canvas context you need an offset so only the shadows (not the text itself) are visible. This point is valid in all cases unless you have one shadow on fully opaque text.

3) Nope, it's only the shadow being rendered to the visible portion of the screen. The text color must be 100% opaque or it will be multiplied against the shadow. It doesn't matter what color as long as it's opaque, so I just chose #000.

I'll try to get back to you with an update soon. I'm reading up on qunit/mocha and looking into how tests are setup (new to me).

btw, "grunt test" command on the homepage docs is returning "Task "test" not found."

mudcube:
Ok, I've made the requested edits, and added a few more fixes, as well as smartly using the renderShadowAlone vs. the way it was before. Let me know if there's anything else you'd like to see changed here.

niklasvh:
There still appears to be some issue with the shadows. On the shadow.html the shadow style is rendered on text that don't have any text applied to the. I've highlighted the issue in the image below:

mudcube:
Woops, I forgot to run clearShadow when textShadow.length === 0 (pushed)

btw, do you think we should alter clearShadow to be the following (if the browser is not being smart then the gaussian blur might still be calculated):

CanvasRenderer.prototype.clearShadow = function() {
    this.setVariable("shadowColor", "transparent").
           setVariable("shadowBlur", 0);
};

I should do a perf, but am running late on getting ready to leave on a trip.

notepad77:
Hello forgive my English , and found a bug in 0.5.0 - alpha2 multiple shadows adding 4 or 5 texts with different shadows and shadow works only in one text

johnTheDudeMan:
If I have 2 different elements that have text-shadowing on, only the first element will render with a text-shadow. Does this issue also affect text-shadowing on multiple elements (not multiple shadowing on a single element)?

PR #508, opened 2015-01-19
Adding ability to set CSS class onto HTML tag of cloned document

cloakedninjas:


niklasvh:
Hi @cloakedninjas!

Thanks for the PR (and +1 for including a test as well!). 0.5 already has an option callback `onclone` which can be used to modify the clone document prior to it being rendered. I've added a test to illustrate how your test case could be implemented using it at:
https://github.com/niklasvh/html2canvas/commit/db887846558c8e6eb0f3194b4ae5aa054f2dcc1a

Do you think that is sufficient to meet your needs?

cloakedninjas:
Ah I didn't see that, so yeah, this will work just as well.

Do you have a full list of options in 0.5 ?

niklasvh:
Unfortunately not yet, will have the documentation updated before the the release of 0.5 though.

PR #545, opened 2015-03-05
Transform matrix for IE

cornedor:
IE10+ returns a matrix3d instead of matrix. Changed the regex to also check for matrix3d and convert matrix3d to a 'normal' matrix

### DIFF
--- a/src/nodecontainer.js
+++ b/src/nodecontainer.js
@@ -258,7 +258,7 @@ NodeContainer.prototype.getValue = function() {
     return value.length === 0 ? (this.node.placeholder || "") : value;
 };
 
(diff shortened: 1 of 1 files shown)


niklasvh:
All the test cases in tests/cases/transform/ return `matrix` for IE11, could you provide an example case where IE returns `matrix3d` where other browsers return `matrix`?

cornedor:
http://jsbin.com/bebuwugeru/1/edit
Will render to:
capture

niklasvh:
Thanks, merged!

usmonster:
Looks like this either breaks a test on Chrome, or the test is flaky:
https://travis-ci.org/niklasvh/html2canvas/builds/53498104#L865

PR #547, opened 2015-03-07
Added the acid2 example and some fixes to render it quite nice

cornedor:
Firefox (left) and Chrome (right).
Acid2 renderd in Firefox (left) and Chrome (right). 

The eye's are done very tricky in acid2 (a object element in a object element in a object element)

I've added support for background-attachment: fixed and a fix for Chrome using 0 as default z-index.


src/nodeparser.js
```diff
@@ -764,6 +764,10 @@ function renderableNode(node) {
 function isPositionedForStacking(container) {
     var position = container.css("position");
     var zIndex = (["absolute", "relative", "fixed"].indexOf(position) !== -1) ? container.css("zIndex") : "auto";
+    // In chrome fixed has a default zIndex of 0 instead of auto
+    if(parseFloat(zIndex, 10) === 0) {
```
niklasvh: `parseFloat` doesn't have a second argument like `parseInt` has, and I believe `parseInt` would be more suitable here anyway.

src/nodecontainer.js
```diff
@@ -164,6 +164,7 @@ NodeContainer.prototype.parseBackgroundSize = function(bounds, image, index) {
 
 NodeContainer.prototype.parseBackgroundPosition = function(bounds, image, index, backgroundSize) {
     var position = this.cssList('backgroundPosition', index);
+    var attachment = this.cssList('backgroundAttachment', index);
```
niklasvh: Could you add some test for `background-attachment` as well, somewhere under https://github.com/niklasvh/html2canvas/tree/master/tests/cases/background

examples/acid2.html
```diff
@@ -0,0 +1,169 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN">
```
niklasvh: The acid2 test got added last week already, under https://github.com/niklasvh/html2canvas/blob/master/tests/cases/acid2.html so I guess this one isn't necessary.

niklasvh:
Thanks for the pull request! I've added a few inline comments.

src/nodeparser.js
cornedor: Meant to use that one indeed :) thanks for pointing out.

cornedor:
I've added the test and removed the acid2 example, while adding the test I found a bug with background position and borders, fixed that too.

NeXTs:
Any updates on this? I do really need support of background-attachment:fixed from this pull request.

cornedor:
I could create a pull request with just the `background-attachment: fixed;` fix.

Issue #385, opened 2014-05-13
bug: fixed html2canvas Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Document': The provided selector is empty. 


fulicat:
just need fix a bug nearby line 1117 in "html2canvas.js" as:

```
function findPseudoElements(el) {
  if(classes.length){ // @@ fixed 
      var els = document.querySelectorAll(classes.join(','));
      for(var i = 0, j = els.length; i < j; i++) {
        createPseudoElements(els[i]);
      }
  }
}
```

herringtown:
@usmonster after a quick look at master, it looks like this is fixed (finding pseudoelements looks like a total re-write along with the rest of 0.5)...

PR #587, opened 2015-05-06
FIXED: https://github.com/niklasvh/html2canvas#issuecomment-52404551

klarkc:


usmonster:
What does this fix? Can you update the description? Thanks!

klarkc:
Sry, so long ago. I don't remember, but It might be related to this: https://github.com/niklasvh/html2canvas/issues/385#issuecomment-52404551

PR #645, opened 2015-08-05
Fix #599

### LINKED ISSUE #599: HTML circle is rendering funny in Chrome/Safari
I have added some circles to the div that I want to send to the canvas with this CSS:
`height: 20px;
  width: 20px;
  border-radius: 20px;
  position: absolute;
  top: -8px;`

http://www.awesomescreenshot.com/image/297705/8c3f811182dc5a909c9c1cbd270dabd2
looks like this.

Works well on Firefox though.

hyojin:
Quick fix for #599

If the border-radius is over half of width/height, it will be set half of width/height.
It prevents circle rendering issue.

### DIFF
--- a/src/nodeparser.js
+++ b/src/nodeparser.js
@@ -617,14 +617,14 @@ function calculateCurvePoints(bounds, borderRadius, borders) {
         width = bounds.width,
         height = bounds.height,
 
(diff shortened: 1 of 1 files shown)


niklasvh:
Thanks, merged!

PR #646, opened 2015-08-05
support for cloning webgl canvas context into 2d canvases

danibuiza:


evgkfx:
Why don't accepted? Html2Canvas still not working with WebGL

niklasvh:
Fixed in https://github.com/niklasvh/html2canvas/commit/250208dc9922e949f78e1b1e624ff92809c2e16a

PR #648, opened 2015-08-06
encode svg data into base64 to prevent some browsers to add html enti…

ncoquelet:
When we add the svg data to img.src, Firefox convert it to replace some characters with html entities and after it can load the image correctly !

Simple poc here : http://jsfiddle.net/3PfcC/

the solution is to convert svg into base64 string and ff can load it without error !

http://jsfiddle.net/0L6mrcqq/


brcontainer:
Use base64 for binary data, for SVG try use `encodeURIComponent`.

Issue #668, opened 2015-08-26
Screenshot from element fail


Thiagocfn:
If call html2canvas to another element than document.body, method fail on html2canvas.js line 602

Thiagocfn:
anything like this will fail:

html2canvas($("#toPrint")., {
                onrendered: function (canvas) {
                    ...
                }
            });

CraigVA:
Try document.getElementById("toPrint") .  jQuery objects aren't always the same as document objects.  Also it looks like you have an extra period before the comma.

Thiagocfn:
Same problem @CraigVA.
This happens, because core.js at line 43: 
var node = ((nodeList === undefined) ? [document.documentElement] : ((nodeList.length) ? nodeList : [nodeList]))[0];

if object is a single element he try get "first element of the list".
I fixed it at #669

Thiagocfn:
@CraigVA  - "Also it looks like you have an extra period before the comma."

Yeah. My mistake on writing this example. But the problem is the same.

PR #669, opened 2015-08-26
Bugfix - when call html2canvas() for an element different from docume…

Thiagocfn:
Fixing issue: Screenshot from element fail #668

### DIFF
--- a/dist/html2canvas.js
+++ b/dist/html2canvas.js
@@ -1963,7 +1963,7 @@ function html2canvas(nodeList, options) {
 
     var node = document.documentElement[0];
     if (nodeList) {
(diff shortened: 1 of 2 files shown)


dist/html2canvas.js
```diff
@@ -1961,7 +1961,10 @@ function html2canvas(nodeList, options) {
         });
     }
 
-    var node = ((nodeList === undefined) ? [document.documentElement] : ((nodeList.length) ? nodeList : [nodeList]))[0];
+    var node = document.documentElement[0];
```
niklasvh: Returns `undefined`

PR #696, opened 2015-10-16
remove fancy font names that canvas appears to ignore

jessecrossley:
The html5 canvas appears to not handle fonts like "Helvetica Neue", so removing any single- or double-quote font names helps ensure that a fallback font is used.


niklasvh:
Should be fixed with https://github.com/niklasvh/html2canvas/commit/77d258f1d8eeccf7e38235463a293b512c1ca3d0

PR #697, opened 2015-10-16
ignore text elements whose parents are svg

jessecrossley:
NodeParser picks up all <text> elements, including those within SVG which are already rendered.  To prevent those text elements from being re-rendered by the conversion process, remove them from the list of text nodes.


niklasvh:
Fixed in 1.0.0

PR #698, opened 2015-10-16
add an options flag 'useBounds' to indicate that the renderer should …

jessecrossley:
Rendering at the exact html pixel size results in blurry images in the canvas.  A workaround was found by 1) creating a canvas with double the bounds of the root html node 2) translating the canvas using the bounds and scaling the canvas by a factor of two 3) setting the canvas's style width and height to the original bounds width and height.  Users may then zoom in to 200% in the browser before the canvas becomes blurry.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -91,7 +91,20 @@ function renderWindow(node, container, options, windowWidth, windowHeight) {
     var bounds = getBounds(node);
     var width = options.type === "view" ? windowWidth : documentWidth(clonedWindow.document);
     var height = options.type === "view" ? windowHeight : documentHeight(clonedWindow.document);
+    if (options.useBounds) {
+        width = bounds.width * 2;
+        height = bounds.height * 2;
+    }
     var renderer = new options.renderer(width, height, imageLoader, options, document);
+    if (options.useBounds) {
+        //scale the canvas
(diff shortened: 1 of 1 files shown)


dahamstor:
Wouldn't it be better to allow the user to specify the scale at which the image is resized? This would allow for screens with 3 DPR such as HTC Butterfly. Then we can detect the currently used DPR on the device (for an example, http://stackoverflow.com/questions/15661339/how-do-i-fix-blurry-text-in-my-html5-canvas) and pass it as the resizeBounds parameter. The code I tried was:

     if (options.resizeBounds != undefined) {
         width = bounds.width * options.resizeBounds;
         height = bounds.height * options.resizeBounds;
     }
     var renderer = new options.renderer(width, height, imageLoader, options, document);
     if (options.resizeBounds ) {
         //scale the canvas
         renderer.canvas.style.width = bounds.width+'px';
         renderer.canvas.style.height = bounds.height+'px';
         renderer.canvas.getContext('2d').translate(-bounds.left * options.resizeBounds, -bounds.top * options.resizeBounds);
         renderer.canvas.getContext('2d').scale(options.resizeBounds, options.resizeBounds);
         // prevents cropping...
         options.canvas = renderer.canvas;
     }

PR #733, opened 2015-11-18
Fix screenshot error if it scroll is not 0, and something.

JixunMoe:
+ Paint to new canvas from point `0, 0` instead.
* Tweak to select surrounding pixels as well;
* Add `window.GM_PROXY` flag to ensure my custom proxy using UserScript(and some hooking) will work.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -117,15 +117,19 @@ function cleanupContainer(container, options) {
 
 function crop(canvas, bounds) {
     var croppedCanvas = document.createElement("canvas");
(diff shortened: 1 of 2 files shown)


JixunMoe:
Fixed the jsonp proxy throw error for `window.html2canvas.proxy` been undefined.

JixunMoe:
Still, not quite sure how the library calculates its offset.

For example, if the element to screenshot is above current view point, it might become nothing (I'm not sure if canvas will store image data in negative axis).

JixunMoe:
Fix too ugly. Close.

PR #747, opened 2015-12-10
A better fix for #697: ignore text elements whose parents are svg

RennieS:
This code will detect whether a text node is within a SVG, no matter how
deep.


chughes87:
+1

niklasvh:
Fixed in 1.0.0

Issue #704, opened 2015-10-21
window.html2canvas is undefined


Bart-Holland:
This is for the Beta2 version!

I downloaded the beta2, did my npm install and then the grunt.
I am now trying to use it in my project (also using require, backbone and jquery).

When running the following script:
```javascript
html2canvas(document.getElementById('sitChartWrap'), {
                allowTaint: true,
                onrendered: function(canvas) {
                    $('#reportEvacChart').html(canvas);
                }
            });
```
I get this error on line 1488:
```javascript
    window.html2canvas is undefined
```

the whole code block in html2canvas:

```javascript
13: [function (require, module, exports) {
            module.exports = function () {
                if (window.html2canvas.logging && window.console && window.console.log) {
                    Function.prototype.bind.call(window.console.log, (window.console)).apply(window.console, [(Date.now() - window.html2canvas.start) + "ms", "html2canvas:"].concat([].slice.call(arguments, 0)));
                }
            };

        }, {}],
```

For now I changed the if statement to:

```javascript
 if (window.html2canvas && window.html2canvas.logging && window.console && window.console.log) {
```
seems to work now, but don't know if it should go like this.

niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/commit/4ebe9c5fcca8fbdb411f7d03c17b8212866fc1ab

PR #754, opened 2015-12-25
not set logging options in window.html2canvas

nishimura-d:
window.html2canvas is undefined when used with RequireJS/AMD.
set in the log module instead.
fixes niklasvh/html2canvas#704.


niklasvh:
Thanks for the PR. This issue has been addressed in https://github.com/niklasvh/html2canvas/commit/4ebe9c5fcca8fbdb411f7d03c17b8212866fc1ab

PR #782, opened 2016-01-28
updated header, added error message.

finetype:
added sane error message for when a bad selector is supplied. Updated the header to say beta4 instead of beta3. Reminified the .min version of the file as well, and also updated its header.


finetype:
Do you know what the webdriver task is?

PR #802, opened 2016-02-19
Force SVGs to be loaded over the proxy, if configured

Paratron:
SVG images would never be loaded via the proxy script and tainted the canvas, even when allowTaint was set to false and a proxy was configured.

I don't know WHY that was the case - however - this fixes the issue.


niklasvh:
Fixed in https://github.com/niklasvh/html2canvas/commit/50608e9cd4327c3503223dca85ab7b2b2b9e629c

PR #804, opened 2016-02-20
fix(proxy): correctly determine supportsCORSImage

alirezamirian:
`crossOrigin` property of image objects can be `null` (which is the default value in FF and Chrome that I tested), and in this case `('crossOrigin' in new Image())` evaluates to true, which doesn't seems to be correct

### DIFF
--- a/src/proxy.js
+++ b/src/proxy.js
@@ -19,7 +19,7 @@ function Proxy(src, proxyUrl, document) {
 var proxyCount = 0;
 
 function ProxyURL(src, proxyUrl, document) {
-    var supportsCORSImage = ('crossOrigin' in new Image());
+    var supportsCORSImage = (new Image()).crossOrigin;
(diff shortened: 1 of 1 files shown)


Siguza:
Google Chrome 50, here on GitHub in the browser console:

`(new Image()).crossOrigin` produces `null`.
`('crossOrigin' in new Image())` produces `true`.

alirezamirian:
Yes, exactly!
Is it true?! I thought in the case that `crossOrigin` evaluates to `null`, `supportsCORSImage` should be `false`. It doesn't work for me when `crossOrigin` property of image object is `null` and we assume browser supports CORS image.

niklasvh:
The `supportsCORSImage` variable defines whether the browser even supports that property, not whether the Image is set to be loaded `crossOrigin` (that image instance is never requested to begin with), so this is working as intended.

PR #851, opened 2016-04-18
Fix Markdown syntax in readme.md

jonathanp:
This commit fixes some markdown syntax errors in readme.md that causes rendering issues on e.g. https://www.npmjs.com/package/html2canvas


niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/commit/50579f399ec8a61144d2bf1061d391ef0af3bceb

Issue #796, opened 2016-02-13
doesn't support font-family property with multiple value


rayax86:
Using 0.5.0-beta4, thanks so much for this great module!

The problem I am having is, say I have a class
`.class {
    font-family: "somefont", "Arial"
}`
where "somefont" doesn't apply, the end result is elements that are .class would be rendered using Arial in the page, but in the canvas the default font is applied instead.

Any idea?

eZ0:
I'm having the same issue.

eZ0:
I found the reason: parent of canvas has to have the font declaration like so:
`.container {
  font-family: Roboto, "Helvetica Neue", sans-serif;
}`

jordan314:
The join/split in CanvasRenderer.prototype.font was stripping out additional fonts. I submitted a fix for it

rsbondi:
I am not sure if this is related, but [here](http://plnkr.co/edit/cQH03OO02BFq8TaxXoNw?p=preview) I think I am seeing the font rendered correctly, but then layered over with an erroneous font?

getJv:
@eZ0 you correct! Thank you!
 As you said, i put `font-family: arial` at canvas parent element and works!. 

Thank you! Thank you!

faiq-sudoers:
Try make the text element like span, h1, h2 ... to display:block

ruslankonev:
not work, I got the same error
![image](https://user-images.githubusercontent.com/2588730/46677010-5e51b900-cbea-11e8-94e2-3cc53a72cbde.png)

rendered by

```
html2canvas(this.$refs.iFr, { useCORS: true }).then(canvas => {
        console.log(canvas)

        document.body.appendChild(canvas)

        canvas.toBlob(function(blob) {
          saveAs(blob, 'Dashboard.png')
        })
      })
```

camohub:
I have the same problem. You can fix it via clone callback and style it separately.

```
               onclone: (document) => {
                        document.querySelectorAll('.button span').forEach(element => {
                            element.style['font-family'] = 'sans-serif';
                        });
		}
```

PR #874, opened 2016-05-12
allow multiple font families

jordan314:
Fix for #796 to allow multiple font families. The join/split was filtering out the other fonts.


robpaveza:
Please accept.

niklasvh:
Should be fixed with https://github.com/niklasvh/html2canvas/commit/77d258f1d8eeccf7e38235463a293b512c1ca3d0

PR #937, opened 2016-09-06
Add condition to use height option to capture entirety of a vertically scrolling element

jzyeezy:
This allows you to capture a screenshot of a vertically scrolling element. Instead of using the generated iFrame's contentWindow height, you can use the provided height option to capture the element's content even if it's hidden from the scrollable view.


ZMDreamer:
I use the  methods you told me, but it didn't work!

Amy0714Angela:
> I use the methods you told me, but it didn't work!

it didn't work,either

Issue #940, opened 2016-09-08
Cropping canvas with incorrect top


Bart-Holland:
I have a div with images which I want to make an image of.
When running html2canvas the top of this div will 'disappear'. 

With running html2canvas in logging mode I found that the cropping in html2canvas is off.

The div itself is positioned at: x = 994 and y = 394. I tested this by: 

```
var e = document.getElementById('colorBlocksEmergency');
var rect = e.getBoundingClientRect();
console.log(rect.x, rect.y);  //994 394
```

However in the logging I see this: 

> 1021ms html2canvas: Cropping canvas at: left: 994 top: 403.83331298828125 width: 40 height: 250
> 1025ms html2canvas: Resulting crop with width 40 and height 250  with x 994 and y 403.83331298828125

I can't seem to figure out why it takes a different y. 
I had slimscroll running, so I stopped using slimscroll but the problem remained.
Also when not using any scrolling the problem was there.

When I test the x and y after the renderDocument, I will also have these 'wrong' x and y. (line 1038)
When testing this before the render document (line 1010) then the x and y are correct.

So somewhere in between the position seems to move...
I can't figure out what goes wrong, so maybe you can find it.

Regards

Bart

Bart-Holland:
So still searching, 
I changed the getBounds method to:

```
 exports.getBounds = function (node) {

                var rect = node.getBoundingClientRect();
                console.log(rect.top, rect.left);

                if (node.getBoundingClientRect) {
                    var clientRect = node.getBoundingClientRect();
                    var width = node.offsetWidth == null ? clientRect.width : node.offsetWidth;
                    return {
                        //top: clientRect.top,
                        top: rect.top,
                        bottom: clientRect.bottom || (clientRect.top + clientRect.height),
                        right: clientRect.left + width,
                        //left: clientRect.left,
                        left: rect.left,
                        width: width,
                        height: node.offsetHeight == null ? clientRect.height : node.offsetHeight
                    };
                }
                return {};
            };
```

to see if this would make a difference, but it did not. 
However the console.log shows every time the method is called. 

In the very beginning after the 'renderDocument' method I can see the values are not correct.
After that, every time the getBounds functionality is called, the values seem correct.

So I looked for the first getBounds after 'renderDocument' and found the 'renderWindow'.
there the values are off as well. And these values are used with cropping. 

Looking at the renderWindow method it seems that the bounds are collected before the rendering is finished. 

```
[...]

PR #941, opened 2016-09-08
Issue #940 cropping area 

Bart-Holland:
Here the changes which fixed issue #940


Bart-Holland:
fails on webdriver
Warning: Task "webdriver" failed. Use --force to continue.

Don't know how to fix this.

PR #980, opened 2016-11-03
Add forceRenderAllText option to help support long strings or words

danbrianwhite:
There are times where the code does not quite understand that when an HTML element has long strings or words without spaces and word wrap is set the text should wrap. It seems better not to try to count and size the letters from the HTML nodes in the Canvas element and allow an override when you know content will be very long. When this situation can occur it would be great to have an option to override the calculation and go straight to force the render. This pull request adds an option to force render all text.


niklasvh:
Will solve this by properly handling unicode and work breaks

PR #1002, opened 2016-11-19
Loosen the test for "text-shadow: 1px 1px 1px"

infinity0:
phantomjs on Debian GNU/Linux gives a slightly different but equivalent result


niklasvh:
`rgb(0,0,0)` is not equivalent to `rgba(0,0,0,0)`

PR #1037, opened 2017-01-15
Fix "Unable to copy canvas content" error when canvas has no 2d context

MinimAI:
I was running into a lot of "Unable to copy canvas content" errors when using html2canvas on a web page containing, for example, a canvas with an experimental-webgl context.

`
canvas.getContext('experimental-webgl');
canvas.getContext('2d'); // now null
newCanvas.getContext('2d');
newCanvas.getContext('experimental-webgl'); // now null
`

This error can be fixed by cloning the canvas using: `clonedCanvas.getContext("2d").drawImage(canvas,0,0);`

Using getImageData is for pixel data access, not for copying canvases. Copying with it is very slow and hard on the browser. It should be avoided.

### DIFF
--- a/dist/html2canvas.js
+++ b/dist/html2canvas.js
@@ -530,17 +530,7 @@ function cloneCanvasContents(canvas, clonedCanvas) {
         if (clonedCanvas) {
             clonedCanvas.width = canvas.width;
             clonedCanvas.height = canvas.height;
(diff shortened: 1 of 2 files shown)


niklasvh:
Fixed in https://github.com/niklasvh/html2canvas/commit/250208dc9922e949f78e1b1e624ff92809c2e16a

PR #1045, opened 2017-01-31
support border radius percent

danomatic:
Adds support for border radius of 50%.  Assumes the percent is based on the element height.  Most useful for border: 50%;

### DIFF
--- a/src/nodeparser.js
+++ b/src/nodeparser.js
@@ -748,10 +748,24 @@ function noLetterSpacing(container) {
 function getBorderRadiusData(container) {
     return ["TopLeft", "TopRight", "BottomRight", "BottomLeft"].map(function(side) {
         var value = container.css('border' + side + 'Radius');
+
         var arr = value.split(" ");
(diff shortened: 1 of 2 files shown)


src/nodeparser.js
```diff
@@ -748,10 +748,24 @@ function noLetterSpacing(container) {
...
         }
+
+        arr = arr.map(function(radius) {
+            if (
+                typeof radius === 'string'
+                && radius.endsWith('%')
+                && container.bounds
+                && container.bounds.height
+            ) {
+                return (parseFloat(radius) / 100) * container.bounds.height;
```
jeffreysnell: If you make this change, this code also handles ovals pretty accurately.

return (parseFloat(radius) / 100) * Math.max(container.bounds.height, container.bounds.width)

niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/commit/adb1f50f0047d6590d768736b9a609e3f37dd599

PR #1067, opened 2017-03-06
support border radius percent

danomatic:
Simple detection and implementation of % in border radius


danomatic:
Oops I already submitted a PR for this...

PR #1068, opened 2017-03-06
fix SVG loading over proxy

danomatic:
Current master fails to load SVG background images across origins.  This fixes that.


niklasvh:
Fixed in https://github.com/niklasvh/html2canvas/commit/50608e9cd4327c3503223dca85ab7b2b2b9e629c

PR #1074, opened 2017-03-10
Fix size bug ff

oizie:
Hi,
this is a fix when an element has width: "auto" or height: "auto" to get the size of the bounds instead of the computed Styles. This would fix some edge case scenarios that happen only in Firefox.


niklasvh:
If this is still relevant, please create an example case where the issue can be duplicated

Issue #876, opened 2016-05-20
If an image is out of the screen, it will not be included, setting height works, but width does not


nigel-dewar:
Hi There,

First of all great work by the way. I have been trying to use your library to take a snapshop of a table. We made it large because otherwise the resolution was poor. 

The problem is the width, anything not on the screen gets cut off when we take a snapshot. The hieght is not an issue though. 
We set the hieght and width to huge values, even if we set width to 100000 it makes no diff
width: 3000,
height: 1500

The snaptshot still gets cut off,

Can you please advise if we are doing some thing wrong or if this is a bug? Any help would be greatly appreciated, I have invested 4 days trying to get this right.

Cheers

niklasvh:
Is this still an issue with [v1.0.0](https://github.com/niklasvh/html2canvas/releases)? If so, could you please share an example on [jsfiddle](http://jsfiddle.net/az5j0766/).

Issue #837, opened 2016-03-30
Only renders what is on shown the screen and the result image is clipped


icakirlar:
This issue can be shown as a duplicate of previous issues #542, #511, #588, etc. However the problem is still open.

Html2canvas is successful when the bounds of the window is never changed and the target element is always %100 visible on the screen. However, this assumption is not always valid. When the bounds of the window is changed or scroll down/up operations only effects the bounds of the defaultView, but the dom structure remains the same after first initiation. In order to fix the problem I applied the following change in the source code and works fine for my case. Hope it can be placed in 0.5.0 release.
- renderDocument expects the windowWidth and windowHeight values. But the actual values (core.js, line 44) refers to document.defaultView.innerWidth and document.defaultView.innerHeight values. If the size of the view changed, unless your element is a part of the actual document, rendering do not take into consideration of your whole element, it only renders the visible area. If you pass appropriate values, it renders the whole element without any problem. 

if you call renderDocument( node.ownerDocument,
                               options,
                               options.innerWidth, //first width of the window when the document is created
                               options.innerHeight,//first heightof the window when the document is created 
                                           index )
like this it works properly.

820fans:
Great !!!  I solved the problem after trying your method . Thank you !!

Dayjo:
@icakirlar awesome, been wondering if this was possible, often my users are using small browser windows and you end up with a cropped off image, when really you want the whole document area whether it's visible or not. 

Any chance of making this a pull request?

NPC:
Hmm, for me this change (unless I did it incorrectly) renders elements that are partially off-screen as completely black, while top-level elements seem to render fine.

Dayjo:
I also have now tried this without success. Trying to make sure the entire 1024 x 576 element is rendered, (if it's clipped off the page / scrolled off the page, that is black) my code is as follows;

```js
            var w = 232, h = 148; var self = this;
                return html2canvas(document.getElementById('da-page-canvas'), {
                    background: '#fff',
                    allowTaint: true,
                    useCORS: true,
                    page: this,
                    onclone: function(doc){
                        var canv = doc.getElementById('da-page-canvas');
                        canv.style.borderColor = 'transparent';
                        canv.className = "da-page-canvas da-screenshot";

                        var selected = doc.querySelectorAll('.ui-selected');
                        for ( var i =0; i< selected.length; i++ ) {
                            selected[i].classList.remove('ui-selected');
                        }
                    },
[...]

Issue #836, opened 2016-03-28
Rendering large page without scroll bars having issue 


NavaneethSankaradevan:
Hi,

I am rendering my html page which is have 540 X 11020 px resolution with IFrame to avoid scrollbars. Actually I am rendering full html page without scrollbar in iframe and sending that to html2canvas. But the generated image is not having full data. I mean it is working properly for 540 X 8000 px (approx). For more than this resolution, I am always getting 540 X 8000 px image. Please suggest me to fix this. 

[http://jsfiddle.net/sodofkcs/35/](url)
Note: For created iframe css won't work in this jsfiddle sample since css file is required

Issue #774, opened 2016-01-25
html2canvas-beta-4 not capturing entire screen


matthew-rister:
I'm been noticing consistent issues (in html2canvas 0.5-beta versions) with screenshot capturing when the page is very long vertically. What appears to happening is that only the portion of the page which is actually within range of the viewport is captured while the rest is cut off.

finetype:
You might try making sure the page is scrolled to the top when calling html2canvas. I have this problem if I put the button at the bottom of the page, but if the button is at the top of the page and the user is scrolled up, all of the page is captured.

TheJson:
I'm using the html2canvas v.0.5.0-beta4 so I don't understand because not capturing full view. It capture only window size. Did you found a solution? Thanks

PR #1083, opened 2017-03-26
Fix rendering of offscreen elements

eKoopmans:
## Bug: Off-screen rendering

Off-screen components are currently not being rendered.

bugfix-offscreen

### Fix

When the `type: 'view'` option is not set, the width of the renderer is set to the `right` and `bottom` bounds of the root node, rather than the document width/height.

#117, #461, #590, #622, #774, #836, #837, #876

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -177,6 +177,9 @@ module.exports = function(grunt) {
     });
 
     grunt.registerTask('webdriver', 'Browser render tests', function(browser, test) {
+        if (process.env.TRAVIS_PULL_REQUEST != "false") {
+            return;
+        }
         var selenium = require("./tests/selenium.js");
         var done = this.async();
         var browsers = (browser) ? [grunt.config.get(this.name + "." + browser)] : _.values(grunt.config.get(this.name));
--- a/src/core.js
+++ b/src/core.js
@@ -87,8 +87,8 @@ function renderWindow(node, container, options, windowWidth, windowHeight) {
     var support = new Support(clonedWindow.document);
     var imageLoader = new ImageLoader(options, support);
     var bounds = getBounds(node);
-    var width = options.type === "view" ? windowWidth : documentWidth(clonedWindow.document);
-    var height = options.type === "view" ? windowHeight : documentHeight(clonedWindow.document);
+    var width = options.type === "view" ? windowWidth : bounds.right;
+    var height = options.type === "view" ? windowHeight : bounds.bottom;
     var renderer = new options.renderer(width, height, imageLoader, options, document);
     var parser = new NodeParser(node, renderer, support, imageLoader, options);
     return parser.ready.then(function() {
--- a/src/core.js
+++ b/src/core.js
@@ -87,8 +87,8 @@ function renderWindow(node, container, options, windowWidth, windowHeight) {
     var support = new Support(clonedWindow.document);
     var imageLoader = new ImageLoader(options, support);
     var bounds = getBounds(node);
-    var width = options.type === "view" ? windowWidth : bounds.right;
-    var height = options.type === "view" ? windowHeight : bounds.bottom;
+    var width = options.type === "view" ? windowWidth : bounds.right + 1;
+    var height = options.type === "view" ? windowHeight : bounds.bottom + 1;
     var renderer = new options.renderer(width, height, imageLoader, options, document);
     var parser = new NodeParser(node, renderer, support, imageLoader, options);
     return parser.ready.then(function() {


Romerski:
@eKoopmans Hi. I tryed your fix but im not able to make this work. I have this: http://i.imgur.com/EE8GyHD.png

Green: Popup when I click print button on Gantt chart.
Blue: HTML gantt chart which I want to render.
Red: What is rendering html2canvas from html (only screen visible part of html)

Is a gantt chart rendered by JQuery.gantt, any help to fix this? Thanks!!

eKoopmans:
Sure @Romerski, I'll take a look. Can you give me a link to your page where you're generating the canvas, or a similar demo?

Romerski:
@eKoopmans Sure, I'll send you an email to erik@erik-koopmans.com with webpage and login user. 
Thanks!!

Romerski:
@eKoopmans Sure! I've sent you an email to erik@erik-koopmans.com. Thank again!

eKoopmans:
@Romerski, I've had a look. The problem is that your element is designed to be exactly the width of its container - so setting `overflow: visible` on the container doesn't solve the problem (which it normally would) because the element isn't any larger than the container.

Here's my fix for your specific case:

1. Before you use `html2canvas` on your element `mod_pf_gantt_244`, set the width of its parent (in your case the `body` of the new window you create) to an arbitrarily large width, something bigger than your element could possibly be:
    ```js
    var el = document.getElementById('mod_pf_gantt_244');
    el.parentElement.style.width = '10000px';
    ```

2. Then, to prevent your element from filling the entire width we just set, change the element's `display` style to `inline-block` (and make sure `width` is set to `auto`):
    ```js
    el.style.display = 'inline-block';
    el.style.width = 'auto';
    ```

That should give your element the right width, so that when you convert to canvas it will display correctly.

Romerski:
Thank you for your reply and your fix! I'll try as soon as I can and I'll give you a response.

Thanks again!!

2017-04-01 20:59 GMT+02:00 Erik Koopmans <notifications@github.com<mailto:notifications@github.com>>:

@Romerski<https://github.com/Romerski>, I've had a look. The problem is that your element is designed to be exactly the width of its container - so setting overflow: visible on the container doesn't solve the problem (which it normally would) because the element isn't any larger than the container.

Here's my fix for your specific case:

  1.  Before you use html2canvas on your element mod_pf_gantt_244, set the width of its parent (in your case the body of the new window you create) to an arbitrarily large width, something bigger than your element could possibly be:

var el = document.getElementById('mod_pf_gantt_244');
el.parentElement.style.width = '10000px';

  2.  Then, to prevent your element from filling the entire width we just set, change the element's display style to inline-block (and make sure width is set to auto):

el.style.display = 'inline-block';
el.style.width = 'auto';

That should give your element the right width, so that when you convert to canvas it will display correctly.

Romerski:
I've just tried and is working like a charm! Thanks again, my problem is fixed now.

I owe you one!

2017-04-03 8:36 GMT+02:00 Yeray Romero <yerayrome@outlook.com<mailto:yerayrome@outlook.com>>:
Thank you for your reply and your fix! I'll try as soon as I can and I'll give you a response.

Thanks again!!

2017-04-01 20:59 GMT+02:00 Erik Koopmans <notifications@github.com<mailto:notifications@github.com>>:

@Romerski<https://github.com/Romerski>, I've had a look. The problem is that your element is designed to be exactly the width of its container - so setting overflow: visible on the container doesn't solve the problem (which it normally would) because the element isn't any larger than the container.

Here's my fix for your specific case:

  1.  Before you use html2canvas on your element mod_pf_gantt_244, set the width of its parent (in your case the body of the new window you create) to an arbitrarily large width, something bigger than your element could possibly be:

var el = document.getElementById('mod_pf_gantt_244');
el.parentElement.style.width = '10000px';

  2.  Then, to prevent your element from filling the entire width we just set, change the element's display style to inline-block (and make sure width is set to auto):

el.style.display = 'inline-block';
el.style.width = 'auto';

That should give your element the right width, so that when you convert to canvas it will display correctly.

eKoopmans:
You're welcome @Romerski, glad it works!

bruce1i:
@eKoopmans thanks, I need the feature and it works for me.

PieterJanVdb:
@eKoopmans Any reason as to why this has not been reviewed / accepted yet?

Darkmift:
Hi,

I am trying to apply a proper print but I am having an issue with the function when called by onclick.

Please see my demonstration:

https://www.youtube.com/watch?v=QrODj-bbv-Y&lc=z23xe15xjvfrgvvvbacdp433tbsjid5ifqxrnt4g5mdw03c010c.1504201255991820

link to project.
https://drive.google.com/drive/folders/0ByKwGLBGU4mPaVUwYVV1QlVQUUE

Any help would be deeply appreciated.

eKoopmans:
@PieterJanVdb the repo owner is doing a big overhaul on html2canvas, I think the new build will have this issue (and many others) resolved. I'm also not sure my solution is perfect...

@Darkmift I've taken a look, and I have to admit I'm confused! I can tell you the different results you're getting are due to your scroll position - so where you say you have to restart the browser, that's not necessary, it's just where you're scrolled to on the screen. But regarding how to fix it... I'm still trying to figure it out!

idolizesc:
Has this been fixed in the latest version?

eKoopmans:
@idolizesc good question! In my testing... yes!! It's fixed!

As of v1.0.0-a1, I believe this issue is resolved, in both foreignObjectRendering and regular rendering. Thanks @niklasvh! I'll close this PR - if anyone has any continued problems with offscreen rendering, please post again here.

OsipovDenis:
@eKoopmans hello!  Thank you for your work! But now, I got the same issue)
I'm upgraded html2canvas to the latest (1.0.0.-alpha.1) version, and it's broke my code.
When I used v.0.5.0 my "point" was in `opacity: 0;` at the container in which was my nodes for transform to canvas.
Now, I try to use anyone of solutions in this repo and it's don't help me.
Could you tell, please:
1) Which difference between 0.5.0 and 1.0.0?
2) Which base rules of rendering offscreen? (if i wanna create image and don't show it process to user?)

Now, it's working correctly if i create image from node on my screen and don't working if i positioned container relative with `left: 2000px;`

ShrutiA11:
when data is too large html2canvas isnt printing the whole data it creates a space after printing certain amount of data else it will be blank page, 
would you pls help me on this

Tholem98:
@ShrutiA11 I had the same problem. I fixed it passing the option scrollY:-window.scrollY
It solve my problem on desktop, but don't solve it on mobile

Issue #978, opened 2016-10-31
I want to show an image that exceeds the area is cut.


dykim1004:
The transform css was using translate, rotate, scale.
JI GIL as a result of its wish to draw on the canvas.

html2canvas 0.5.0-alpha1 

-- result --

![html2canvas 0 5 0-alpha1](https://cloud.githubusercontent.com/assets/16068092/19855274/f6d97fee-9fb6-11e6-871c-29ad6762ef0d.png)

custom souce ... 

CanvasRenderer.prototype.clip = function(shapes, callback, context) {
    //this.ctx.save();
    shapes.filter(hasEntries).forEach(function(shape) {
        this.shape(shape).clip();
    }, this);
    callback.call(context);
    //this.ctx.restore();
};

-- result --

![customimg](https://cloud.githubusercontent.com/assets/16068092/19855340/58381e76-9fb7-11e6-831f-11261515e81f.png)

I do wonder sikijineun edits an error occurred in the utility function.
Otherwise, when the corresponding overflow hidden in one version, the car I want to provide an option to be able to cut img screent shot. Thank you

niklasvh:
Is this still an issue with [v1.0.0](https://github.com/niklasvh/html2canvas/releases)? If so, could you please share an example on [jsfiddle](http://jsfiddle.net/az5j0766/).

Issue #510, opened 2015-01-19
Box-shadow and CSS transforms skew not supported? Fiddle provided


samsonkhisty:
Greetings, 
http://jsfiddle.net/fuggfuggfugg/q572g6q0/4/ - This is my working example. 
Tried searching but was unable to find anything with regards to skewed lines.

samsonkhisty:
updated: http://jsfiddle.net/fuggfuggfugg/q572g6q0/5/

samsonkhisty:
partly working code: http://codepen.io/fuggfuggfugg/pen/QwvvVq

usmonster:
Follow #221 for box shadow.

samsonkhisty:
Thanks. 
Another update, this code works for skew. http://codepen.io/fuggfuggfugg/pen/ZYyBRv
My first example used :before & :after which probably isn't supported.

usmonster:
@samsonkhisty What browser(s) are you testing in? IE, for example, apparently doesn't support pseudoelements in its implementation of `.getComputedStyle` until IE11... Though I see the same issue you're having on Chrome, so you're probably right that it's something not completely implemented in html2canvas..

samsonkhisty:
* My very first codepen (http://jsfiddle.net/fuggfuggfugg/q572g6q0/4/) does not work in any browser 
  * transform:skew specified in :before/:after is not applied 
  * Although gradients specified applied in the generated canvas 

* The latest codepen (http://codepen.io/fuggfuggfugg/pen/ZYyBRv) works perfectly in Chrome, Safari and FF. 
  * IE9: The HTML/CSS is perfect but canvas output on IE9 shows square boxes. 
  * I haven't had the opportunity to test in any latest IE or Opera yet.

nyroDev:
I'm also experiencing this issue with css transform.
I'm using a css similar to http://codepen.io/AtomicNoggin/pen/fEish

After digging a little in the code, I have the feeling that it has something to do with the clip definition of the rotated element, but not sure to get the whole.picture.
If the main dev could give me a hint on how to solve this problem, I could give it a try.

usmonster:
@samsonkhisty I just got a chance to check out your codepen example, and it looks fine in IE11 in both "Edge" and IE10 modes. IE9 mode throws some JS errors, though they seem to be coming from codepen itself, not html2canvas. According to [their documentation page](http://blog.codepen.io/documentation/), they only support IE10+. Can you try with jsFiddle or jsbin?

[...]

Issue #349, opened 2014-03-16
Not hiding overflow:hidden data


ankur1708:
I am using html2canvas to make 1 image out of 3-4 images. All these images can be zoomed for which i am using jquery zoom. There is one parent div holding all these images.
Overflow is hidden for all these images so that user can set it as per his wish.
Without zooming, on Html2Canvas, i am getting a perfect canvas element. But when any image is zoomed, and then i try HTML2Canvas, the hidden content is shown in canvas created. Is there a solution for this which i may not aware of.
My functionality is similar to http://www.fotor.com/features/collage.html. 
Thanks in Advance.

niklasvh:
Is this still an issue with [v1.0.0](https://github.com/niklasvh/html2canvas/releases)? If so, could you please share an example on [jsfiddle](http://jsfiddle.net/az5j0766/).

Lxiaozhuang:
请问  这个解决了么  ？  我现在也遇到了这个问题  scale放大后的图片最后在html2canvas生成图片后 隐藏的部分都显示出来了

brcontainer:
@Lxiaozhuang In html2canvas-0.4.1 or in html2canvas-1.0?

nishantgupta11:
-> Add CSS **overflow-y: scroll**; to HTML element **pdfHTML**
  -> Add attribute **width** & **windowWidth** to html2canvas method, which should be equal to width of **pdfHTML** Div
  
   
    html2canvas(document.querySelector(".pdfHTML"), {
        width: 1350, windowWidth: 1350
    }).then(canvas => {
        base64stringpdf = canvas.toDataURL("image/jpeg");
    });

PR #1084, opened 2017-03-26
Fix clipping on css-transformed elements

eKoopmans:
## Bug: Clipping on CSS-transformed elements

Elements with `transform` CSS apply a transformation directly to the canvas, adjusting all canvas actions. This causes incorrect `clip` bounds, since the transformations are not considered when calculating bounds.

bugfix-clip-transform

### Fix

Any canvas `transform` is temporarily undone while applying clips, then re-applied before rendering the element. This involved giving `clip` access to the `container` object to test for transforms.

#220, #349, #510, #978

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -177,6 +177,9 @@ module.exports = function(grunt) {
     });
 
     grunt.registerTask('webdriver', 'Browser render tests', function(browser, test) {
+        if (process.env.TRAVIS_PULL_REQUEST != "false") {
+            return;
+        }
         var selenium = require("./tests/selenium.js");
         var done = this.async();
         var browsers = (browser) ? [grunt.config.get(this.name + "." + browser)] : _.values(grunt.config.get(this.name));
--- a/src/nodecontainer.js
+++ b/src/nodecontainer.js
@@ -240,6 +240,11 @@ NodeContainer.prototype.parseTransformMatrix = function() {
     return this.transformMatrix;
 };
 
+NodeContainer.prototype.inverseTransform = function() {
+    var transformData = this.parseTransform();
+    return { origin: transformData.origin, matrix: matrixInverse(transformData.matrix) };
+};
+
 NodeContainer.prototype.parseBounds = function() {
     return this.bounds || (this.bounds = this.hasTransform() ? offsetBounds(this.node) : getBounds(this.node));
 };
@@ -279,6 +284,14 @@ function parseMatrix(match) {
         });
         return [matrix3d[0], matrix3d[1], matrix3d[4], matrix3d[5], matrix3d[12], matrix3d[13]];
     }
+}
+
+function matrixInverse(m) {
+    // This is programmed specifically for transform matrices, which have a fixed structure.
+    var a = m[0], b = m[2], c = m[4], d = m[1], e = m[3], f = m[5];
+    var det = a*e - b*d;
+    var M = [e, -d, -b, a, b*f-c*e, c*d-a*f].map(function(val) { return val/det; });
+    return M;
 }
 
 function isPercentage(value) {
--- a/src/nodeparser.js
+++ b/src/nodeparser.js
@@ -328,11 +328,11 @@ NodeParser.prototype.paintElement = function(container) {
     var bounds = container.parseBounds();
     this.renderer.clip(container.backgroundClip, function() {
         this.renderer.renderBackground(container, bounds, container.borders.borders.map(getWidth));
-    }, this);
+    }, this, container);
 
     this.renderer.clip(container.clip, function() {
         this.renderer.renderBorders(container.borders.borders);
-    }, this);
+    }, this, container);
 
     this.renderer.clip(container.backgroundClip, function() {
         switch (container.node.nodeName) {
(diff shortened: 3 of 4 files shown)


oizie:
This would a good PR. Plus the webdriver task it's being skipped and at least is the first PR sucessful since long time ago.

haxxxton:
@eKoopmans, just implemented this in my own fork, and found that it caused text to not render. The  included examples basically rendered as if the the text had a transparent colour. 
Let me know if this is not the case for you and ill see what i can do to create a replicable example :)

eKoopmans:
Hi @haxxxton, definitely interesting, I wasn't able to reproduce though: [Fiddle](https://jsfiddle.net/m8tjpqcf/) (text shows up fine for me in this).

You'll notice [here](https://github.com/niklasvh/html2canvas/pull/1084/files#diff-1e3122b763426ab76f7d89d0d688f151R460) that for `paintText` I send `container.parent` to `clip`, so it looks like that parent is then having trouble. `.hasTransform` calls `.parseTransformMatrix` which calls `.prefixedCss` which calls `.css`, and when it gets there it needs to be an Element for `getComputedStyle`.

All that said, it should be sufficient to change [line 72 of canvas.js](https://github.com/niklasvh/html2canvas/pull/1084/files#diff-e711d3178639416c00742138c3223c08R72) to `if (container instanceof Element && container.hasTransform()) {`, but I'm curious what context it's getting a container that isn't an Element.

haxxxton:
@eKoopmans, huge apologies about this.. as this isnt a pull request on a repo im in control of, i had to copy-paste the changes across, and i missed that subtle difference of `container.parent` over `container`. Changing it to the _correct_ code fixes the comment i raised. Again, sorry for wasting your time :)

If you're interested in an npm accessible version of html2canvas that has a number of these PR's merged, you can have a look [here](https://www.npmjs.com/package/html2canvas-fixes).

eKoopmans:
Hey @haxxxton, no worries, glad it was a simple fix. Thanks for the npm link!

gharwood1:
Hey @eKoopmans 

I believe this is causing a issue with:
```
.transform {
  -webkit-transform: rotate(-60deg);
  transform: rotate(-60deg);
}
```
Check out this [fiddle](https://jsfiddle.net/aqxhraw3/)

Thanks for your work!

eKoopmans:
Hi @gharwood1, you're right that rotations still aren't handled correctly by this pull request (though it's still an improvement) - see issue #1138 for a bit more info, specifically [this fiddle](https://jsfiddle.net/jjarrpw7/) illustrates the problem. I've also created a [modified version of your fiddle](https://jsfiddle.net/ubg018cv/) using divs, again to illustrate that clipping isn't being done correctly.

On Chrome at least, the fiddle you posted illustrates a second issue, which is that spans (and anything that's not `display: block`) shouldn't have any transforms applied to them (info [here](https://stackoverflow.com/questions/14883250/css-transform-doesnt-work-on-inline-elements)). So even though you've applied the rotate style, the original `span` appears un-rotated, but html2canvas rotates it. Definitely interesting!

Back to the original clipping issue: As I mentioned at #1138, fixing the clipping issue on rotated spans would require a much more serious rewrite of how clipping is being done in html2canvas. Right now html2canvas stores a stack of bounds information for all of an element's ancestors, and when drawing each element it applies the full stack of clips based on those bounds. Since the rotated element needs a rotated clip, either the bounds need to be stored in a rotated form and *all further bounds need to have that rotation as part of their bounds calculation*, or (better IMO) clipping and transforms can be applied as html2canvas descends through its children, so it can be applied correctly once and used for all children.

eKoopmans:
**Important update:** This PR introduces new issues for certain CSS transformations and should be used with caution (see [my last comment](https://github.com/niklasvh/html2canvas/pull/1084#issuecomment-316738589) for more info). I apologize for any problems it may have caused.

@niklasvh, what are your thoughts on applying each node's clipping to the canvas as you descend down the tree, and restoring the canvas to its previous state on the way back out, rather than the current method of keeping an array of clipping bounds?

niklasvh:
@eKoopmans I had a [stab at overflows ](https://github.com/niklasvh/html2canvas/commit/f278ba4f2221fb229a2b323c4716d169b5607964)last night (as well as with transforms). They certainly are a bit problematic. 

I think we cannot just apply the clips as we descend the tree as the drawing order of the canvas is based on stacking contexts instead of the DOM tree order (which overflows use). One alternative I thought about was applying transforms on a node level instead of globally for the whole stack and descendants, that way you could apply the clips along with the transforms together, in the right order. I've yet to try it out, and probably will first focus on getting the v1.0.0 branch up to par with the current functionality before I continue working on this.

eKoopmans:
Right, fair enough. The other alternative could be to keep things as is but apply transforms to the clip vectors... Not sure if that would cause other problems.

niklasvh:
That could be an alternative as well, but may get quite tricky with border-radius bezier curves etc?

eKoopmans:
Yeah that was what stopped me when I started going down that path originally. I don't understand all the bezier calculations, especially with the inner and outer borders.

niklasvh:
Will need to think about this a bit more, but for now focusing on getting the remaining features in first.

eKoopmans:
This PR still needs work (see [comment](https://github.com/niklasvh/html2canvas/pull/1084#issuecomment-319736849) and #1138). The issue isn't yet resolved in v1.0.0-a1 when using non-foreignObject rendering.

Issue #838, opened 2016-03-30
Not Rendering ligatures


chadwtkns:
Hi Im using html2canvas to generate a jpeg but "ff", "fl", and "fi"s are not rendering. See images below. Any reason why or has anyone else encountered this?

<img width="171" alt="screen shot 2016-03-30 at 11 20 53 am" src="https://cloud.githubusercontent.com/assets/5611546/14151115/a0bdca30-f669-11e5-919d-8b0ccd992c45.png">
<img width="217" alt="screen shot 2016-03-30 at 11 21 00 am" src="https://cloud.githubusercontent.com/assets/5611546/14151116/a0c0a732-f669-11e5-9eda-9125c7c90c2f.png">

floDroidCoder:
Hi, I've also encountered the same issue.
I'm using the v0.4.1 - 7.9.2013 version of html2canvas and solved it by replacing line 1214 :
      textList = (!options.letterRendering && /^(left|right|justify|auto)$/.test(textAlign) && noLetterSpacing(getCSS(el, "letterSpacing"))) ?

with
      textList = (!options.letterRendering && /^(left|right|justify|auto**|start|end|center**)$/.test(textAlign) && noLetterSpacing(getCSS(el, "letterSpacing"))) ?

gurigurico:
This doesn't work if css letter spacing is applied on the text...

burtontanner:
Could anyone tell me the css that I would need to remove to get this to work for me as well? @mtskf

burtontanner:
if i change this:

textList = (!options.letterRendering && /^(left|right|justify|auto)$/.test(textAlign) && noLetterSpacing(getCSS(el, "letterSpacing"))) ?
      textNode.nodeValue.split(/(\b| )/)
      : textNode.nodeValue.split("");
to this:

      textList = true ?
      textNode.nodeValue.split(/(\b| )/)
      : textNode.nodeValue.split("");

and that seemed to work for me

cscouto:
I solve this problem using a condition to check if the word has "fi"
` textList = ((!options.letterRendering && /^(left|right|justify|auto)$/.test(textAlign) && noLetterSpacing(getCSS(el, "letterSpacing"))) || (textNode.nodeValue.indexOf("fi") !== -1) ) ?`

eKoopmans:
### The problem

[...]

PR #1085, opened 2017-03-26
Fix ligatures when letter-rendering

eKoopmans:
## Bug: Ligatures

Text is sometimes rendered incorrectly, when the font uses [ligatures](https://en.wikipedia.org/wiki/Typographic_ligature#Stylistic_ligatures) and the text is rendered letter-by-letter. Adjacent letters (i.e. 'fi') are rendered on-top of each other.

bugfix-ligatures

### Fix

When text is being rendered per-character (instead of per-word), this fix adds the `font-variant-ligatures: none` CSS to the parent element. There is also a fix to the `.font()` handling of `font-variant` CSS data, as not all `font-variant` values are supported on canvases.

#838


eKoopmans:
Tested, this is still an issue in v1.0.0-a1 (using non-foreignObjectRendering). I can work out a fix for the latest build if you like.

Mylogin-info:
The "f" and "l" are overlapping. 

overlap

Is there going to be a fix soon?

PythonDream:
Hi, i tried using the fix from [838](https://github.com/niklasvh/html2canvas/issues/838)
but with no success.

Is there gonna be an update soon that fixes this?

Issue #1088, opened 2017-03-27
Build testing: Broken webdriver task


eKoopmans:
There's a bug in the testing process with the `webdriver` task - nobody's pull requests are passing. I've tracked down the problem to `initBrowser` in `tests/utils.js` - the browser is never initialized (the `.init(settings)` promise never resolves). I believe that the selenium server that is needed to perform the tests is not being started.

The browser object is created using [wd](https://www.npmjs.com/package/wd), which explains that:

> WD.js does not start the selenium server. You may use the [selenium-standalone](https://www.npmjs.com/package/selenium-standalone) package to install and start a selenium server.

#### Starting the selenium server

I've tried installing and running `selenium-standalone` on my machine, and after removing `port: 4445` (line 35 of `utils.js`), I got the webdriver test to get a bit further - several browsers pop up on my machine before the test fails.

`html2canvas` tests using **Travis**, so I'm guessing [this guide](https://docs.travis-ci.com/user/gui-and-headless-browsers/) is relevant. This would make sense since `utils.js` uses the environment variables `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` as described in those instructions. Where I'm stuck is where they describe [starting a web server](https://docs.travis-ci.com/user/gui-and-headless-browsers/), which I believe isn't happening in our tests. Just not sure how to set that up here.

It's a shame, I think many/most current PRs are failing just because of this problem!

### Related issues/pull requests

#782, #941, #1084, #1085, #1086, #1087

eKoopmans:
I found the problem - the `webdriver` task can't be performed on pull requests, because we don't have access to the Sauce Labs environment variables. This is by design - those variables are meant to be kept private - so the only solution is to bypass the `webdriver` task on pull requests.

#1089 solves the problem. If you have made a pull request and want it to pass the tests, I recommend rebasing off of that branch. Here's what you would do:

    git fetch https://github.com/eKoopmans/html2canvas.git bugfix/webdriver
    git checkout your-branch
    git rebase bugfix/webdriver

That will place the `bugfix/webdriver` commit before your branch, which should fix the issue. You can then update your repo on GitHub:

    git push -f

niklasvh:
Moving onwards, the whole testing infrastructure should be rethought a bit. I would hope to move to more accurate and consistent automated tests and limit the webdriver tests purely just for parsing (as the rendering should be consistent if the parsed values are the same across browsers).

niklasvh:
Fixed

PR #1087, opened 2017-03-26
Add dpi/scale options for custom resolution

eKoopmans:
## Feature: Custom resolution with DPI/scale options

Simple support for increasing/decreasing the canvas resolution with the new `dpi` and `scale` options.

feature-scale-dpi

### Implementation

When a custom `scale` is set, the canvas' width/height are multiplied by that scale while keeping its CSS width/height at the original. Then `ctx.scale` is used to scale all future canvas actions (see [here](http://stackoverflow.com/a/26047748/4080966) for more info).

The option `{dpi: 96}` is equivalent to `{scale: 1}`, and larger values of either option will increase the resolution. If both options are present, `scale` is ignored in favour of `dpi`.

### Related issues/pull requests

#127, #176, #241, #373, #379, #390, #621, #842, #900, #947, #1039, #1057

PR #1086, opened 2017-03-26
Add box-shadow rendering

eKoopmans:
## Feature: Box-shadows

You know 'em, you love 'em: `box-shadow`s! Requests for this feature date back to May 2013.

feature-box-shadow

### Implementation

In `paintElement`, after rendering backgrounds but before borders, there is now `renderShadows`. It uses a new `.mask()` tool which behaves the opposite of `.clip()`, meaning the shadow will be drawn only outside the element container.

This feature includes support of multiple `box-shadow`s.

### Related issues/pull requests

#220, #221, #246, #510, #597, #673, #763

PR #1089, opened 2017-03-27
Remove grunt webdriver task for pull requests

eKoopmans:
## Bug: Grunt webdriver task for pull requests

The `webdriver` task in the Grunt testing process doesn't work for external pull requests, and is causing all PRs to fail.

### Fix

The webdriver task is set up to connect to [Sauce Labs](https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-Sauce-Labs), which hosts the Selenium cloud for browser testing. To connect to Sauce Labs, `tests/utils.js` looks for the environment variables `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY`.

Unfortunately, those variables are [encrypted in .travis.yml](https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml) (as they should be), which means that they are *not available to pull requests from forks*. Thus the test fails, because it never connects to Sauce Labs.

To prevent this guaranteed failure, I have disabled the `webdriver` task if it detects that it is being performed in a pull request, as recommended in the [Travis documentation](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions):

> To work around this, restrict these tests only to situations where the environment variables are available, or disable them for pull requests entirely.

#782, #941, #1084, #1085, #1086, #1087, #1088

(And really every open pull request, since none can currently pass with the current setup).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -177,6 +177,9 @@ module.exports = function(grunt) {
     });
 
     grunt.registerTask('webdriver', 'Browser render tests', function(browser, test) {
+        if (process.env.TRAVIS_PULL_REQUEST != "false") {
+            return;
+        }
         var selenium = require("./tests/selenium.js");
         var done = this.async();
         var browsers = (browser) ? [grunt.config.get(this.name + "." + browser)] : _.values(grunt.config.get(this.name));


niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/blob/v1.0.0/karma.conf.js#L6

Issue #390, opened 2014-05-24
Canvas rendering is blurry on retina displays


adilapapaya:
Thanks for the great library! One thing I'm noticing though is that the screenshots are blurry when using a retina display. For example, here's what the screenshot of [https://github.com/niklasvh/html2canvas](github.com/niklasvh/html2canvas) looks like on a retina display (as taken from [the test page](http://html2canvas.hertzen.com/screenshots.html)).
![screen shot 2014-05-24 at 1 29 51 pm](https://cloud.githubusercontent.com/assets/4053020/3075582/3f716922-e382-11e3-9d34-98a60ea9e2ad.png)

I think maybe there needs to be a check for the ratio between the `window.devicePixelRatio` and the `context.backingStorePixelRatio` somewhere in the rendering of the canvas...?

p.s. Here are some useful links I came across while trying to get to the root of the issue. 
- [Apple's Safari Canvas Guide](https://developer.apple.com/library/safari/documentation/audiovideo/conceptual/html-canvas-guide/SettingUptheCanvas/SettingUptheCanvas.html).
- [html5rocks.com HiDPI Blog Post](http://www.html5rocks.com/en/tutorials/canvas/hidpi/).
- [Related question on Stack Overflow](http://stackoverflow.com/questions/19674262/html5-canvas-render-to-texture-and-hidpi-display-appears-blurry).

syxchina:
not fixed

YPCrumble:
:+1:

btm1:
+1

retep007:
+1 scalling and zooming css styles are not working right ?

breim:
+1 please fix

zqcloveping:
+1 fix

breim:
@zqcloveping 
Bro, try this configs, not solve 100% but increases quality

html2canvas(document.getElementById('flag'), {
				useCORS: true,
				allowTaint: true,
				letterRendering: true,
				onrendered: function(canvas) {
					var ctx = canvas.getContext('2d');
					ctx.webkitImageSmoothingEnabled = false;
					ctx.mozImageSmoothingEnabled = false;
					ctx.imageSmoothingEnabled = false;
				},
			});

zqcloveping:
@breim 
Bro，3Q for you help.But it seems to be less obvious.Rendering problems in retina devices

vsesh:
Please fix it!

nicholasaleks:
+1 please fix

[...]

Issue #135, opened 2012-11-09
No support for Retina display on iPhone?


Adc100:
On iPhone and Safari I'm using html2canvas to take a screenshot of a div with text and images.  It works but it's very blurry due to the retina display on iPhones.  This problem is usually easily corrected for images and canvases by using the technique below but I can't get it to work with html2canvas.  I've been modifying the canvas.width, height, and ctx.scale in Canvas.js but it's always blurry.

This works on all other canvases but not with html2canvas
var canvas = document.getElementById('pic');
canvas.width = 640;
canvas.height = 960;
canvas.style.width = "320px";
canvas.style.height = "480px";

For more information on canvas and retina see
http://zsprawl.com/iOS/2012/03/html5-canvas-and-retina-displays/
and
http://blog.iwalt.com/2010/08/generating-highresolution-graphics-with-html5s-canvas-element.html

usmonster:
@niklasvh can this be closed in favor of #390?

PR #1090, opened 2017-03-28
fix retina displays blurry

xyxu:
#390 #135


niklasvh:
Fixed in https://github.com/niklasvh/html2canvas/tree/v1.0.0

PR #1121, opened 2017-05-09
Markdown requires a space after the hashtags

ferrybig:
If this space is missing, the markdown may get wrongly rendered on platforms like github


eKoopmans:
Agreed, good improvement to the readme.

eKoopmans:
Actually, [this resource](https://en.support.wordpress.com/markdown-quick-reference/) suggests that the closing hash marks should have a space separating them as well, and [this editor](http://dillinger.io/) confirms that. So I recommend adding those spaces too.

eKoopmans:
Recommend closing in favour of #1140.

PR #1153, opened 2017-06-28
Fix blurry images

### LINKED ISSUE #576: Blurry image fix
Recently had an issue with images exporting and looking incredibly blurry. 

Looking at the html2canvas logs I saw the x position and crop position were returning as 174.5. This half a pixel was causing the images to render with an excessive amount of blur. 

To force this to stop (as I couldn't think of a reason this would be the desired outcome) I have added **Math.floor()** to the **top, bottom, right, left** positions.  (I did not have any issues with width and height.). 

``` php
    function getBounds (node) {
        if (node.getBoundingClientRect) {
            var clientRect = node.getBoundingClientRect();
            var width = node.offsetWidth == null ? clientRect.width : node.offsetWidth;
            return {
                top   : Math.floor(clientRect.top),
                bottom: Math.floor(clientRect.bottom || (clientRect.top + clientRect.height)),
                right : Math.floor(clientRect.left + width),
                left  : Math.floor(clientRect.left),
                width : width,
                height: node.offsetHeight == null ? clientRect.height : node.offsetHeight
            };
        }
        return {};
    }
```

CKGrafico:
Fix #576 from @graham73may that never did the PR, don't know why.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -177,6 +177,9 @@ module.exports = function(grunt) {
     });
 
     grunt.registerTask('webdriver', 'Browser render tests', function(browser, test) {
+        if (process.env.TRAVIS_PULL_REQUEST != "false") {
+            return;
+        }
         var selenium = require("./tests/selenium.js");
         var done = this.async();
         var browsers = (browser) ? [grunt.config.get(this.name + "." + browser)] : _.values(grunt.config.get(this.name));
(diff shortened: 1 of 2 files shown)


g5codyswartz:
I would love to see this merged! : )

niklasvh:
The blurryness has been fixed in 1.0.0

PR #1177, opened 2017-08-04
Supports Base64 encoding images CORS

yb121:
The mobile ios v10.3. X images are configured across domains to work properly


niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/commit/8999c7618184e1564d19d9c367a488c22c0d65a3

Issue #858, opened 2016-04-21
Autofocus attribute should be removed when cloning nodes


efbiaiinzinz:
When current page that is being cloned contains an element which has autofocus="true" attribute, the rendering process pulls the focus away from whatever element was active, which changes the current UI.
Isue appears in Chrome, was not able to reproduce in Firefox. Patches cloneNode function manually with
if (clone.removeAttribute) clone.removeAttribute('autofocus');
And the focus-loss issue went away.

PR #1252, opened 2017-10-25
add tests for autofocus fix

bgschiller:
I wrote some tests for #858, as I was experiencing the same issue. Then I noticed it was already fixed. But still, they're good tests Bront.


niklasvh:
Thanks for the PR! Sorry for not replying earlier, will try to get this merged asap!

PR #1268, opened 2017-11-16
fixing loading images in iframes when using foreignObject

jolan78:
probably necessary too for other elements inside iframes

### DIFF
--- a/src/Clone.js
+++ b/src/Clone.js
@@ -264,6 +264,21 @@ export class DocumentCloner {
                 case 'SELECT':
                     clone.value = node.value;
                     break;
+                case 'IMG':
+                    var dt = node.ownerDocument.doctype;
+                    if (
+                        dt == undefined ||
+                        dt == 'null' ||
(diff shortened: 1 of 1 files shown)


jolan78:
npm test returns no errors on my side using MacOS 10.11.6 and the same browser versions as travis

niklasvh:
Do you have a test case where the issue can be replicated? Would like to make sure any fixes are not later broken again by adding tests for them.

jolan78:
here is a simple test : 
[test.zip](https://github.com/niklasvh/html2canvas/files/1495566/test.zip)
it will not work for me if I simply open it with a browser due to cross-origin issues.

niklasvh:
Interesting if the doctype actually affects the rendering, in that case the cloned document should actually use the same doctype.

jolan78:
I will continue to discuss the doctype issue in https://github.com/niklasvh/html2canvas/issues/1298 . this pull request was about loading images in iframes, sorry for the confusion

PR #1291, opened 2017-12-06
Scale foreignObject image to fill the canvas

eKoopmans:
Fix scaling on foreignObject rendering.

### DIFF
--- a/src/renderer/ForeignObjectRenderer.js
+++ b/src/renderer/ForeignObjectRenderer.js
@@ -39,7 +39,7 @@ export default class ForeignObjectRenderer {
                 );
             }
 
-            this.ctx.drawImage(img, -options.x * options.scale, -options.y * options.scale);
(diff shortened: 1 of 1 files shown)


niklasvh:
This change breaks all the tests under tests/reftests/options. The scaling should already be correct as the SVG image is created with the scaling (https://github.com/niklasvh/html2canvas/blob/master/src/renderer/ForeignObjectRenderer.js#L24-L27). Do you have an example where it isn't working correctly?

eKoopmans:
Hmm interesting, it's scaled wrong for me. Yeah, I'll make a fiddle if you can help out with the external resource.

eKoopmans:
[Fiddle](https://jsfiddle.net/0k4o7vks/1/). Tested in Chrome (FF, IE, and Edge all fall back to canvas rendering instead of foreignObject). The entire div is half its intended width in the canvas on my machine.

niklasvh:
Yep, seeing same results as you. Need to figure out what is breaking the tests vs. what is breaking this.

PR #1292, opened 2017-12-06
Bugfix: Fix underlines

eKoopmans:
Fix underline positioning.

Since `ctx.textBaseline` is set to 'bottom', underlines need to be positioned relative to the bottom of the text, not the top.

### DIFF
--- a/src/renderer/CanvasRenderer.js
+++ b/src/renderer/CanvasRenderer.js
@@ -212,7 +212,7 @@ export default class CanvasRenderer implements RenderTarget<HTMLCanvasElement> {
                             const {baseline} = this.options.fontMetrics.getMetrics(font);
                             this.rectangle(
                                 text.bounds.left,
-                                Math.round(text.bounds.top + baseline),
+                                Math.round(text.bounds.top + text.bounds.height - baseline),
                                 text.bounds.width,
                                 1,
                                 textDecorationColor


niklasvh:
This change makes the underlines appear at the top of the text, at least with the examples in text/underline.html. Do you have an example where this change yields better results?

eKoopmans:
Hmm, yeah the underlines were at the top on my test, I'll duplicate. Quick question for you, do you have any suggestion what I can link to as an external resource for jsFiddle?

niklasvh:
By external resource do you mean an image?

eKoopmans:
Yeah, the v1.0.0-alpha1 build. I'm used to using rawgit + `dist`, and it won't accept the file at `releases/download` either. I feel like I must be missing an obvious option but I don't know what it is (short of hosting it myself, which I guess is an option)!

niklasvh:
You can hotlink from releases https://github.com/niklasvh/html2canvas/releases

eKoopmans:
You're right! I swear I tried before and it wasn't working, my bad.

eKoopmans:
[Fiddle](https://jsfiddle.net/8bfLnnqp/2/). Tested in Chrome, FF, IE, and Edge, all have underlines above the text on my machine.

niklasvh:
Interesting... I'm seeing the same results there. Can you check what you see for text/underline.html?

eKoopmans:
Maybe a stupid question, but... How do I check text/underline.html? Opening it directly gives me `html2canvas is not defined`, and running the npm tests passed but doesn't give me much to look at...

niklasvh:
Run `npm start` and go to http://localhost:8080

eKoopmans:
Thanks. So you're right, `underline.html` renders correctly - **but** when I rerun html2canvas in the console and append it to document.body, it's rendered wrong. So something about how html2canvas is being invoked...

niklasvh:
You are correct! Still trying to understand why that is the case though

eKoopmans:
Hey thanks! Did you sort out the issue in the testers?

niklasvh:
Nope, haven't figured out why they render wrongly yet.

eKoopmans:
Not sure why, but commit 38749bc, "Fix canvas rendering on Chrome", also fixed underline rendering (on all browsers). Since it was part of the same release as my contribution, the effect is that my commit *re-broke* the underlines. I've tested and it was definitely 38749bc that caused the fix.

@niklasvh, definitely best to revert my commit 0c8d38d. Any idea why 38749bc would have fixed the underlines?

PR #1293, opened 2017-12-06
Bugfix: Fix parseLetterSpacing for decimal values

eKoopmans:
parseLetterSpacing was returning zero for decimal values less than 1.

This was causing TextBounds to not use letterRendering on elements with non-zero letter-spacing.


niklasvh:
Good find! I fixed the issue that was failing the tests (parseFloat only takes 1 argument) and added an example of this use case to the tests (https://github.com/niklasvh/html2canvas/commit/166cbba7c2f1a89ecd3eacc72e8576d566de299d).

eKoopmans:
Oops! Thanks for the fix.

PR #1305, opened 2017-12-11
Added support for radial gradients

### LINKED ISSUE #808: Radial Gradient not working in background
Hi there.

What a nice librairy!!

But I have a problem rendering radial gradient. I checked in the code and the case "radial-gradient" is actually missing.

I have to implement radial gradient there.

Any idea how to implement this?

matthias-christen:
**Summary**

This PR adds support for radial gradients.
I also added a function (+tests) to convert old the old webkit format to the W3C format, so tests/reftests/background/radial-gradient.html renders correctly. I also added some more tests (tests/reftests/background/radial-gradient2.html.

I realize that this is quite a large pull request, hopefully it is possible to include somehow :-)

One thing I wasn't quite sure about (mainly concerning the type system) is how to represent radial gradients. I renamed the type `Gradient` to `LinearGradient` and added a new one, `RadialGradient`. The `parseGradient` function returns the union of both types. Maybe it would be cleaner to define a base type `Gradient` and let `LinearGradient` and `RadialGradient` extend that.

I pulled parsing color stops into its own function, and I also refactored the conversion of old "from" / "to" color stops a bit and added support for parsing the old "color-stop" function.

---

* [x] Support for radial gradients (including definitions in the old webkit format)

---

No radial gradients were rendered previously.
Fixes #808 

---

**Test plan (required)**

* Existing (mostly using old formats): tests/reftests/background/radial-gradient.html
* New (for testing some more properties, using the W3C format): tests/reftests/background/radial-gradient2.html

### DIFF
--- a/src/Gradient.js
+++ b/src/Gradient.js
@@ -3,14 +3,22 @@
 
 import type {BackgroundSource} from './parsing/background';
 import type {Bounds} from './Bounds';
+import NodeContainer from './NodeContainer';
 import {parseAngle} from './Angle';
 import Color from './Color';
-import Length, {LENGTH_TYPE} from './Length';
+import Length, {LENGTH_TYPE, calculateLengthFromValueWithUnit} from './Length';
+import {distance} from './Util';
 
 const SIDE_OR_CORNER = /^(to )?(left|top|right|bottom)( (left|top|right|bottom))?$/i;
 const PERCENTAGE_ANGLES = /^([+-]?\d*\.?\d+)% ([+-]?\d*\.?\d+)%$/i;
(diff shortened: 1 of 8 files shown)


niklasvh:
Nice work! I'll take a closer look tonight, but from a quick glance few notes (which I can fix):
 - Make `Gradient` an interface that Linear and Radial gradients implement
 - Use existing `Vector` instead of `Point`
 - Update https://github.com/niklasvh/html2canvas/blob/master/docs/features.md

matthias-christen:
Thanks! If there is anything I can help you with, please let me know!

niklasvh:
Merged

PR #1320, opened 2017-12-15
Added support for rendering enumerations in ordered lists and for list-style

### LINKED ISSUE #177: Bullets are not rendered
If i have an unordered or ordered list in my html markup, the bullets are not rendered by html2canvas. I think this is quite an important issue because this is a standard feature of html supported in all browsers since a long time ago.

### LINKED ISSUE #308: List style images are missing
Try to put this page into the test console:
 http://www.w3schools.com/cssref/trycss_list-style-image.htm

Bullets wont show up

matthias-christen:
**Summary**

This PR adds support for rendering the enumerations (bullets, numbers, etc.) in ordered and unordered lists and for list style images.
For the actual formatting, an external package is used (https://www.npmjs.com/package/liststyletype-formatter), I hope that's OK.

* [x] Feature: support for rendering enumerations in (un)ordered lists (list-style-type, list-style-position, list-style-image)

More accurate rendering of (un)ordered lists.

**Test plan (required)**

Added tests/reftests/list/liststyle.html, which tests some more list styles in addition to the already existing tests, in the tests/reftests/list folder.

**Closing issues**

Fixes #177 
Fixes #308

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
   "homepage": "https://html2canvas.hertzen.com",
   "license": "MIT",
   "dependencies": {
-    "punycode": "2.1.0"
+    "punycode": "2.1.0",
+    "liststyletype-formatter": "latest"
   }
 }
--- a/src/ListItem.js
+++ b/src/ListItem.js
@@ -1,0 +1,146 @@
+/* @flow */
+'use strict';
+
+import type {BackgroundSource} from './parsing/background';
+import type ResourceLoader from './ResourceLoader';
+
+import {parseBackgroundImage} from './parsing/background';
+import {copyCSSStyles, getParentOfType} from './Util';
+import NodeContainer from './NodeContainer';
+import TextContainer from './TextContainer';
+import ListStyleTypeFormatter from 'liststyletype-formatter';
+
+// Margin between the enumeration and the list item content
+const MARGIN_RIGHT = 7;
+
+export const LIST_STYLE_POSITION = {
+    INSIDE: 0,
+    OUTSIDE: 1
+};
+
+export type ListStylePosition = $Values<typeof LIST_STYLE_POSITION>;
+
+export type ListStyle = {
+    listStyleType: string,
+    listStyleImage: BackgroundSource,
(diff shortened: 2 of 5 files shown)


niklasvh:
Thanks for the PR! I started looking through this over the weekend.

Few things that need to be resolved (which I started fixing already):
 - The list style should not be tied to `LI` tagName but instead `display: list-item` as a `div` or any other element can have an list item just as long as its `display` is `list-item` same way a `LI` won't have a list item if you change it's `display`.
 - Supporting all the different variants of list styles without any build time options to selectively choose which ones/languages to support might increase the package size excessively without much gain. Same applies with changes to word-break etc support for different languages where I was planning on tackling that issue. Ancient versions of html2canvas had list style support, with a few of the variants supported: https://github.com/niklasvh/html2canvas/blob/0.4.1/src/Generate.js#L391
 - Move the list style parsing out of ListItem and into parsing/listStyle

As said I am working on fixing those issues already so no need to change the PR.

andersborgabiro:
At least I'd be happy to be able to apply at least unordered bullets at all, without too much "magic", as in my case a user edits content via TinyMCE. It's possible to apply custom styles to tags, as long as "li" is used (as that's what a user chooses in the toolbar) rather than something else completely.

niklasvh:
All types of list-style are supported for now, smaller subsets with specific languages will be worked with the word breaking and other language dependent features.

andersborgabiro:
Great! Something fun to do during the holidays :).

tininess:
> All types of list-style are supported for now, smaller subsets with specific languages will be worked with the word breaking and other language dependent features.

I using the newest pre-release but not work :( https://jsfiddle.net/boLxkgj8/265/

RUSSCITY:
> > All types of list-style are supported for now, smaller subsets with specific languages will be worked with the word breaking and other language dependent features.
> 
> I using the newest pre-release but not work :( https://jsfiddle.net/boLxkgj8/265/

ok, you was not using the latest version, now it looks like it works:
https://jsfiddle.net/jeavhg05/4/

PR #1321, opened 2017-12-15
Improved support of "content" for pseudo-elements

### LINKED ISSUE #971: Pseudo Element Content Attribute Not Rendering in Non Webkit Browsers
First off, thank you for the really awesome plugin. It really helped me with my project!

However, I found an issue in it that with the script. and a solution. I'm not familiar with GitHub so perhaps this would be a "Pull Request" instead, but I'm putting it as an "Issue" since it's currently an issue.

In your `NodeParser.prototype.getPseudoElement` function of html2canvas.js (line 2057), the script goes through pseudo elements and uses their styled content as content to print. This works for things like `el:after { content: 'hello world' };` but not for things like `el:after { content: attr(title) }`.

You can see it at https://jsfiddle.net/whtsavpp/5/. Chrome 52 renders it to say 'title' while Firefox 49.0.1, and IE 11 renders it to be 'attr(title)'.

This appears to happen in non-Webkit browsers, as other browser renders content according to the style sheet while Webkit renders according to the value.

We fixed it by adding the following lines to the function, after line 2062:

`var content = stripQuotes(style.content);
 var matches = style.content.match(/^attr\(.*\)$/);
    if(matches) {
        var attr = style.content.replace(/^attr\((.*)\)$/, '$1');
        content = container.node.getAttribute(attr);
    }`

If somebody can find an alternative solution, that would work too, but this is how we fixed the problem.

I have attached my revised version of the file as reference.
[html2canvas.txt](https://github.com/niklasvh/html2canvas/files/553811/html2canvas.txt)

Thank you.

matthias-christen:
**Summary**

This PR improves the support for the `content` style declaration for pseudo-elements.

The following was added to the existing support of strings and `url()`:
* [x] Handling multiple content items (`content` can take a list of whitespace-separated items)
* [x] Support for the `attr()` function
* [x] Support for counters (`counter()` and `counters()`)
* [x] Support for quotes (`open-quote`, `close-quote`)

Better support for `content`.
This PR is not yet exhaustive, though (i.e., `content` still supports a wider range of item types).

**Test plan (required)**

* Added tests/reftests/pseudo-content.html
* Added a mocha test, tests/node/pseudonodecontent.js, to test the content parser.

**Closing issues**
Fixes #971

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -22,7 +22,7 @@ Below is a list of all the supported CSS properties and values.
    - border-width
  - bottom
  - box-sizing
- - content (**Does not support `attr()`**)
+ - content
  - color
  - display
  - flex
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
   "homepage": "https://html2canvas.hertzen.com",
   "license": "MIT",
   "dependencies": {
-    "punycode": "2.1.0"
+    "punycode": "2.1.0",
+    "liststyletype-formatter": "latest"
   }
 }
--- a/src/Clone.js
+++ b/src/Clone.js
@@ -394,7 +394,7 @@ const inlinePseudoElement = (
     node: HTMLElement,
     clone: HTMLElement,
     style: CSSStyleDeclaration,
-    contentItems: Array<PseudoContentItem>,
+    contentItems: ?Array<PseudoContentItem>,
     pseudoElt: ':before' | ':after'
 ): ?HTMLElement => {
     if (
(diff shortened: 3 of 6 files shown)


niklasvh:
Thanks for the PR! With the list-style PR merged now, will start taking a look at this over the weekend :)

matthias-christen:
Cool! And thanks for merging and working on the list-style PR!
This PR uses the liststyletype-formatter package again for the counters, but that should be easily replaceable with your implementation.

niklasvh:
There are still some issues with all versions of IE, but they are non breaking issues (it just doesn't parse the counters etc correctly), so will merge this already.

matthias-christen:
Thanks for merging! I'll have a look at the IE issues.

Issue #1364, opened 2018-01-04
text-decoration underline is misplaced


FDIM:
### Bug reports:

http://jsfiddle.net/8rjbyqgj/

As you can see in jsfiddle, links that have a `text-decoration: underline` rule renders with a line above it

### Specifications:

 * html2canvas version tested with: 1.0.0-alpha-8
 * Browser & version: chrome 63 & IE 11
 * Operating system: Windows 7

niklasvh:
Related: https://github.com/niklasvh/html2canvas/pull/1292

FDIM:
So I've looked at the values before rectangle / line is rendered and if I set baseline to 0, it renders correctly (or close enough).
https://github.com/niklasvh/html2canvas/blob/b7c7464c5f19404959ea169046c5fd1b1ac357ed/src/renderer/CanvasRenderer.js#L253

I am not certain about details, but could the fix be `Math.round(text.bounds.top + baseline)` ? bounds.height was 19 and baseline value was 17 in my case

eKoopmans:
Hey @FDIM, thanks for finding the problem, you're right that 0c8d38d ended up causing the very error it was meant to fix. I explained a bit more [here](https://github.com/niklasvh/html2canvas/pull/1292#issuecomment-360700136).

edguz2408:
I recently started using the last version of HTML2Canvas which should include the fix described above, however I kept getting a misplaced underline when I used text-decoration: underline just like it's mentioned at the beginning of this thread, I managed to fix the issue by modifying Math.round(text.bounds.top + text.bounds.height - baseline) to Math.round(text.bounds.top + text.bounds.height) in CanvasRenderer.js file

Issue #1363, opened 2018-01-04
version is logged to console


FDIM:
https://github.com/niklasvh/html2canvas/blob/b7c7464c5f19404959ea169046c5fd1b1ac357ed/src/index.js#L38

could the above line use the logger and not console directly? 

I can submit a PR if you'd like to :)

mc-suchecki:
I was thinking about PR to that, when I saw your issue I decided to do it. 👍

niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/commit/da2794f7f773e9f4a18f6b00ba5129988610e81a

PR #1394, opened 2018-01-15
Bugfix for underline position and console log

FDIM:
I've removed unnecessary console log that logs version (even incorrectly) at runtime and fixed position of link underline. The latter can be reproduced in examples/demo2.html

I've changed logger a bit as well, so that it would use console directly and not via window. Tests did not pass otherwise and complained that window is undefined.

Fixes for #1364 and #1363

### DIFF
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -18,11 +18,13 @@ export default class Logger {
 
     // eslint-disable-next-line flowtype/no-weak-types
     log(...args: any) {
-        if (this.enabled && window.console && window.console.log) {
+        // eslint-disable-next-line no-console
+        if (this.enabled && typeof console !== undefined && console.log) {
             Function.prototype.bind
-                .call(window.console.log, window.console)
+                // eslint-disable-next-line no-console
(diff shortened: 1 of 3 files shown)


FDIM:
```
15 01 2018 20:30:37.775:ERROR [launcher]: Cannot start Chrome
	[4687:4687:0115/203037.649789:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/google/chrome/chrome-sandbox is owned by root and has mode 4755.
```

Tests are passing locally for me on windows machine in chrome and firefox

eKoopmans:
Hi @FDIM, you may want to submit the two pieces (underline and console) as separate PRs. You're right that the underline fix was causing problems, thanks for catching it - I've submitted a PR to revert that change.

FDIM:
@eKoopmans I've opened another PR with changes related to console

Issue #1374, opened 2018-01-08
IE 11 issue, Error: Uncaught (in promise): Error: Member not found.


Prabhat2404:
Got error : "html2canvas $npm_package_version. Error: Uncaught (in promise): Error: Member not found." 
When i try to get canvas.

Code:
 html2canvas(document.querySelector("#capture")).then(function (canvas) {
            console.log(canvas);
            document.getElementById('here-appear-theimages').appendChild(canvas);
        });
Using:
Angular version : 4.0.1
OS: Window 10
Also added refrence of: https://cdn.jsdelivr.net/npm/es6-promise/dist/es6-promise.auto.js
Attached error screenshot.

![error_ie_11](https://user-images.githubusercontent.com/16957593/34688771-3fd75ba8-f4d9-11e7-91ae-669fcb875197.png)

jelkins8:
I added a link to the following promise library and it seemed to fix IE 11 for us.
https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.3.4/bluebird.min.js

niklasvh:
Without code to reproduce the error I can't really help.

Prabhat2404:
This is demo sample i have hosted on GitHub you can see - [GitHub Html2Canvas Demo](https://prabhat2404.github.io/#/pages/dashboard)

Full code is also uploaded on github, if you want to see: [Code package](https://github.com/Prabhat2404/prabhat2404.github.io/blob/master/ng2-admin-with-html2canvas.tgz)

How to run project:
1) Unzip package
2) Run command: "npm install" from directory to install dependency 
3) To run project command: "npm start"

Prabhat2404:
Hi @jelkins8,

I have tried this solution but its not working for me.
This is demo sample i have hosted on GitHub you can see - [GitHub Html2Canvas Demo](https://prabhat2404.github.io/#/pages/dashboard)

Thanks

simonpinn:
Hi @niklasvh , 
I too have tried with es6-promise and bluebird without success in IE11, can confirm that [GitHub Html2Canvas Demo](https://prabhat2404.github.io/#/pages/dashboard) also fails in IE for me.
I receive ```html2canvas: Error: Member not found.``` followed by ```Possible Unhandled Promise Rejection: Error: Member not found.```
thanks!

GPranali:
I am facing the same issue for html2canvas: member not found for IE.
For chrome it is working perfectly fine.
[...]

PR #1415, opened 2018-01-24
IE11 Member not found fix Wrap accesing the cssText property in a try...catch

simonpinn:
https://github.com/niklasvh/html2canvas/issues/1374

Reproduced the above issue and located the cause, it happens in the createElementClone function when it attempts to access the cssText property, for some reason when using font-awesome, an element with fa-spin throws this error in IE when accessing the property.

Wrapping the property access in a try catch solves the immediate issue.

### DIFF
--- a/src/Clone.js
+++ b/src/Clone.js
@@ -230,19 +230,17 @@ export class DocumentCloner {
         }
 
         if (node instanceof HTMLStyleElement && node.sheet && node.sheet.cssRules) {
-            const css = [].slice
-                .call(node.sheet.cssRules, 0)
-                .reduce((css, rule) => {
-                    try {
-                        if (rule && rule.cssText) {
-                          return css + rule.cssText;
-                        }
-                        return css;
-                      } catch (err) {
-                        this.logger.log('Unable to access cssText property', rule.name);
-                        return css;
-                      }
-                });
+            const css = [].slice.call(node.sheet.cssRules, 0).reduce((css, rule) => {
+                try {
+                    if (rule && rule.cssText) {
+                        return css + rule.cssText;
+                    }
+                    return css;
+                } catch (err) {
+                    this.logger.log('Unable to access cssText property', rule.name);
+                    return css;
+                }
+            });
             const style = node.cloneNode(false);
             style.textContent = css;
             return style;
--- a/src/Clone.js
+++ b/src/Clone.js
@@ -232,7 +232,17 @@ export class DocumentCloner {
         if (node instanceof HTMLStyleElement && node.sheet && node.sheet.cssRules) {
             const css = [].slice
                 .call(node.sheet.cssRules, 0)
-                .reduce((css, rule) => css + rule.cssText, '');
+                .reduce((css, rule) => {
+                    try {
+                        if (rule && rule.cssText) {
+                          return css + rule.cssText;
+                        }
+                        return css;
+                      } catch (err) {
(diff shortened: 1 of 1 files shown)


davesmith9088:
I was having the same error in IE11, until I saw this issue and the try..catch worked for me as well.  I wasn't using font-awesome, not sure exactly what was triggering it.

akalinux:
So when will the next release be out?  I am having the same issue with IE11

simonpinn:
@akalinux - I've asked @niklasvh  to have a look at a regression that causes a failing rendering test https://github.com/niklasvh/html2canvas/issues/1374 - unless you are familiar with the unit tests?

NielsSkovgaard:
Your code gave various Console errors in IE11, however this doesn't:

~~~~
            if (node instanceof HTMLStyleElement && node.sheet && node.sheet.cssRules) {
                let self = this;
                var css = [].slice.call(node.sheet.cssRules, 0).reduce(function (css, rule) {
                    try {
                        if (rule && rule.cssText) {
                            return css + rule.cssText;
                        }
                        return css;
                    } catch (err) {
                        self.logger.log('Unable to access cssText property', rule.name);
                        return css;
                    }
                }, '');
                var style = node.cloneNode(false);
                style.textContent = css;
                return style;
            }
~~~~

Now, the above will log the following in the Console i.e. there seems to be a problem with the Flatpickr (date picker) in my project, which is solved by the try...catch above:

html2canvas: Unable to access cssText property fpFadeInDown

cyouden:
I think you just need to add an empty string as your initial value on the call to reduce (according to MDN: "Calling reduce() on an empty array without an initial value is an error."). The call to reduce in the original code has that, and trying to use this patch locally in a project throws exceptions until I added that. Tests seem to be passing locally as well with that addition.

xiaocoati:
@simonpinn Hello! I use the latest version 1.0.0-rc.0, but also has the same problem in IE11.

niklasvh:
@xiaocoati can you share an example that replicates it on jsfiddle?

xiaocoati:
> @xiaocoati can you share an example that replicates it on jsfiddle?

Hello niklasvh! Because I'm in China, I can't open jsfiddle, so I make an example on JSRUN [https://jsrun.net/uJXKp](url)

xiaocoati:
@niklasvh I have make an example used version 1.0.0-rc.1, but still has the problem in IE11. The attached ZIP file is my demo, you can download it, run `npm install`,
 `npm run mock` and `npm run dev` . Can you please have a look? Really apreaciate!
[demo.zip](https://github.com/niklasvh/html2canvas/files/3080187/demo.zip)

wugel:
@niklasvh I have the sam problem using version 1.0.0-rc.1 in IE11. 
Thank you for help

xszh:
@niklasvh I also meet the probelm using version 1.0.0-rc.1 in IE 11. 
When checking the PRs of src/Clone.js, there was a merge #1693 about 21days ago. It looks like the try catch now will throw everything except "SecurityError". So the exception throw by IE(member not found) will be throw again and then everything stoped. 
In my case, the exception is thrown by the element which is a svg gen by font-awesome.
Would you take a look? Thanks a lot

rappad:
I am having the same problem as @xszh on IE 11

after throwing this exception, the lib stops the process

`     if (e.name !== 'SecurityError') {
                this.logger.log(e);
                throw e;
            }
`

PR #1425, opened 2018-01-31
Bugfix/no console log

### LINKED ISSUE #1363: version is logged to console

FDIM:
I've removed unnecessary console log that logs version (even incorrectly) at runtime.

I've changed logger a bit as well, so that it would use console directly and not via window. Tests did not pass otherwise and complained that window is undefined.

Fixes #1363

### DIFF
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -18,11 +18,13 @@ export default class Logger {
 
     // eslint-disable-next-line flowtype/no-weak-types
     log(...args: any) {
-        if (this.enabled && window.console && window.console.log) {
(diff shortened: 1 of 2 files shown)


niklasvh:
Fixed with https://github.com/niklasvh/html2canvas/commit/da2794f7f773e9f4a18f6b00ba5129988610e81a

Issue #1108, opened 2017-04-23
html2canvas does not support -webkit-text-stroke?


jonatanavital:
I'm using this css:
    -webkit-text-stroke-width: 1px;
    -webkit-text-stroke-color: white;
But it does not render the text with it.
Any ideas?

kunal886496:
He does not care

ivictbor:
Thanks a lot @arzynik , I've published your branch to npm: https://www.npmjs.com/package/html2canvas-stroke , it unblocked https://github.com/devforth/painterro text-stroke feature (though html2canvas-stroke is very updated, probably someone has time capacity to merge a new changes or needs this, poke me)

KupinFB:
Hey, I forked the branch 1.0.0-rc.7 and implemented -webkit-text-stroke-width, -webkit-text-stroke-color, paint-order supporting (based on https://github.com/niklasvh/html2canvas/pull/1540).
Link: https://github.com/KupinFB/html2canvas
Also I've published it to be able to use.
Link: https://www.npmjs.com/package/html2canvas-stroke2
Can i keep it?

PR #1540, opened 2018-05-24
Text-stroke and Paint-order support

### LINKED ISSUE #759: Is there no support for -webkit-text-stroke
I like html2canvas.But it looks like it's not supporting text stroke.Is there any way i could find a solution for this?

spacedevin:
This ads support for `-webkit-text-stroke`, `-webkit-text-stroke-width`, `-webkit-text-stroke-color`, and `paint-order`. 

Note that even through these are webkit prefixes, Firefox still uses them.

`paint-order` is required in order to get strokes to render as an outline of the text and not in the middle.

`paint-order` does not currently work fully in chrome, as chrome reads the value of the property, but does not apply it in the dom. Firefox and Safari work just fine.

See `tests/text/stroke.html` for an example

---

Fixes #759 and #1108

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -12,9 +12,9 @@ Below is a list of all the supported CSS properties and values.
        - url()
        - linear-gradient()
        - radial-gradient()
-   - background-origin 
+   - background-origin
    - background-position
-   - background-size   
+   - background-size
  - border
    - border-color
    - border-radius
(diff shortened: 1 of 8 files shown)


ivictbor:
Thanks a lot @arzynik , I've published your branch to npm: https://www.npmjs.com/package/html2canvas-stroke , it unblocked https://github.com/devforth/painterro text-stroke feature (though html2canvas-stroke is very updated, probably someone has time capacity to merge a new changes or needs this, poke me)

PR #1571, opened 2018-06-22
Attempt to include flex-wrap and justify-content

Bishares:
I tried to include flex-wrap and justify-content from flexbox. npm test was sucessfull.
It doesn't seem to change the behaviour of the canvas drawing. Maybe I forgot something?
these css properties should be easy to include because its naturally supported by chrome.


niklasvh:
These properties impact layout, which this library does not perform. It relies on the browser to compute layout, so parsing this information will not affect the render result in any way.

PR #1649, opened 2018-09-07
Fix CSS content property for Safari

### LINKED ISSUE #1610: [Safari 10.1.2] Can't render FontAwesome Icon in canvas.
### Bug reports:
I am using this library to convert FontAwesome Icons on the html page into canvas, then converting them to base64 string, and finally, to byte[] for use in a crystal report. On Firefox and Chrome, this functionality works perfectly. On Safari 10.1.2, the canvas element doesn't render the icon; It shows as a blank, white square.

To Reproduce:
1. Add a fontawesome 4 icon on to html page.
2. Call 
`html2canvas(document.querySelector(`#icon`), { logging: true, allowTaint: true })`
3. When promise fulfills, call:
```
    canvas.toDataURL();
    document.body.appendChild(canvas);
```
The canvas element shows up as blank white. The base64 string is valid, but it is also blank white (as expected i suppose).

On a separate, but less important note. I had a bug in safari only, where if the <i> element has display:inline-block, the canvas would be empty. I had to change it to display:inline-flex for the canvas to load properly-ish

### Specifications:

 * html2canvas version tested with: v1.0.0-alpha.12
 * Browser & version: Safari 10.1.2
 * Operating system: OSX Yosemite 10.10.5

Dragooon:
Safari does not pass content strings in double quotes, so they get ignored. This breaks packages such as Font-Awesome on Safari and iOS.

Fixes #1610

### DIFF
--- a/src/PseudoNodeContent.js
+++ b/src/PseudoNodeContent.js
@@ -311,6 +311,8 @@ const addOtherToken = (tokens: Array<Token>, identifier: string): void => {
         case 'close-quote':
             tokens.push({type: TOKEN_TYPE.CLOSEQUOTE});
             break;
+        default:
+            tokens.push({type: TOKEN_TYPE.STRING, value: identifier});
     }
 };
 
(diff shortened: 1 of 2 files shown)


WriterStat:
Be advised #1649 pull request did not fix the problem for #1736.
@mjsarfatti

Fontawesome icons still show up as blank in Apple Safari after applying the above pull request in our local code.

So fix still needed for this. Best,

PR #1677, opened 2018-10-05
Handle ligatures

### LINKED ISSUE #1676: Problem with Ligatures in Textareas

### Bug reports:

Some browsers/versions seem to have problems with ligatures in textareas / text-input fields, so that the letters of the ligatures get rendered over each other instead of next to each other.

I encountered this problem first on Android with Chrome version ~ 65.x. Note that more current versions of Chrome for Android (e.g. 69.x) do not have this problem anymore. So this problem is potentially present for (Android) Chrome versions <= 65.x (the example screenshot below was made using Chrome/WebKit 58.x).

This [jsfiddle](https://jsfiddle.net/jgvpzyex/4/) contains a basic example for testing this, but it would need to be run e.g. as Cordova app on Android:  
the "create" button will use `html2canvas` creating an image that is then appended at the bottom -- if the ligature problem is present, the text of the text-area would show the ligatures conflated, as visible in this screenshot (see 2nd last textarea):

--------

ligature-problem_screenshot_2018-10-05-17-08-23

--------

### Specifications:

 * html2canvas version tested with: v1.0.0-alpha.12
 * Browser & version: Chrome/WebKit 58.0.3029.83
 * Operating system: Android 5.1.1

russaa:
**Summary**

* [x] handle ligatures in text-fields

As explained in issue #1676 `html2canvas` does not handle ligatures correctly for some (older) versions of Chrome/WebKit for text in text-fields (namely `<input>` and ´<textarea>´):
the letters of the ligatures are rendered unto each other instead of next to each other.

The underlying reason being that when the text-boundaries for individual letters are queried, these older Chrome versions return the boundaries for the whole ligature, so each letter of the ligature starts at the same coordinates.

The PR uses an heuristic for detecting these conflated text-boundaries and then distributes the letters with equal distance within the boundary of the complete ligature.

Example (Android 5.1.1, Chrome/WebKit 58.0.3029.83):
ligature_fix

As can be seen, the "fix" does not actually render the ligatures, but the individual letters next to each, which in some cases would look just same as the rendered ligatures but in some cases it may look as if rendered with disabled ligatures (see examples that have a space after the ligature).
 
Since it is an heuristic, the provided solution requires an extra option `fixLigatures` to be set to `true`; otherwise the heuristic will not be used (i.e. works the same way as without this PR), e.g.
```javascript
html2canvas(elem, {fixLigatures: true})
```

There is a similar PR [pull/1085](https://github.com/niklasvh/html2canvas/pull/1085) which tries to solve the issue by setting the CCS style `font-variant-ligatures: none`; however this seems to reset the `font` styles completely, so it did not work in the circumstances described here.

**Test plan (required)**

added template `tests/reftests/options/fixLigatures.html`

See also issue #1676:  
this problem is not present in all browsers; one way of replicating it would be to use the example provided in this [jsfiddle](https://jsfiddle.net/jgvpzyex/4/) and wrap it in a [Cordova](https://cordova.apache.org/) app with platform `android` (tested on a device with Android version 5.1.1).

**Code formatting**

* [x] Please make sure that code adheres to the project code formatting. Running `npm run format` will automatically format your code correctly.

**Closing issues**

Fixes #1676

_related issues that might be effected: #838 #1541_


rememberlenny:
Any reason this isn't merged? @niklasvh

hasanparasteh (approved):
Looks great! Please consider testing more fonts like `FiraCode`, `Iosevka` and etc...

antoine10245:
Hey can we have this merged too?
I see multiple PRs adressing major issues are not merged

PR #1681, opened 2018-10-11
fixes #1593

### LINKED ISSUE #1593: ERROR Error: Uncaught (in promise): [object Event]
Im using this plugin angular 5 and I encounter this error every time when I take screenshot. Please suggest how to handle in Typescript not in Jquery all.
Thanks

adamszeibert:
Angular `ngZone` throws an error due to an uncaught `Promise.reject()`. This PR adds the necessary `catch`

### DIFF
--- a/src/Clone.js
+++ b/src/Clone.js
@@ -225,7 +225,10 @@ export class DocumentCloner {
                                 );
                             }
                         })
-                );
+                )
+                .catch(() => {
+                    // avoid Angular ngZone error
+                });
             return tempIframe;
         }
 


Warriormuffin:
What needs to be fixed for this to go in?

adamszeibert:
@Warriormuffin if you use html2canvas in an Angular project then ngZone throws an error due to an uncaught promise rejection. This can freeze the app for a couple of seconds or in some cases can render the app useless until reload.

The root cause of the issue is that `getIframeDocumentElement` returns `Promise.reject()` in some cases. This PR adds a `catch()` to this promise, which does nothing, but it solves the Angular ngZone issue.

adamszeibert:
@Warriormuffin or are you referring to the failed CI build? I have no idea why that's failing.

Warriormuffin:
Yup, I am hoping to get the fix in the application I am currently working
on!

adamszeibert:
@Warriormuffin you can get the fixed dist from here https://pastebin.com/iR4kC1vP or you can checkout my branch and build your own.

niklasvh:
This should be fixed in v1.1.1

PR #1693, opened 2018-10-29
Wrap .sheet.cssRules access in try...catch.

### LINKED ISSUE #1673: node.sheet.cssRules accessed outside of try-catch, so the thrown DOMException is uncaught
### Bug reports:

In Clone.js (line 232), in the createElementClone function, the code checks if the node is an instance of HTMLStyleElement and then reads the sheet.cssRules property. However, if the style links to an external style sheet, a DOMException will be thrown. This check needs to be inside a try-catch statement. Since it isn't, the exception is thrown all the way and the screen capture fails.

### Specifications:

 * html2canvas version tested with: 1.0.0-alpha12
 * Browser & version: Firefox 60.2.1esr
 * Operating system: Windows 7

PaulJThompson:
**Summary**

Wrap accesses to .sheet.cssRules in a try...catch block. On Firefox, this access can cause a DOMException to be thrown ('SecurityError') which interrupts the entire process. Log-and-resume to avoid it.

* [node.sheet.cssRules accessed outside of try-catch, so the thrown DOMException is uncaught #1673](https://github.com/niklasvh/html2canvas/issues/1673) Bug 1

The uncaught exception means that the entire html2canvas operation is interrupted and no image is produced. At least for our application, the cssRule can be safely ignored.

**Test plan (required)**

I'm still working on this. Adding a file into tests/reftests isn't working because the documents don't get cloned first before copying them.

**Code formatting**

**Closing issues**
closes #1673

### DIFF
--- a/src/Clone.js
+++ b/src/Clone.js
@@ -229,22 +229,32 @@ export class DocumentCloner {
             return tempIframe;
         }
 
-        if (node instanceof HTMLStyleElement && node.sheet && node.sheet.cssRules) {
-            const css = [].slice.call(node.sheet.cssRules, 0).reduce((css, rule) => {
-                try {
-                    if (rule && rule.cssText) {
-                        return css + rule.cssText;
-                    }
-                    return css;
-                } catch (err) {
-                    this.logger.log('Unable to access cssText property', rule.name);
-                    return css;
-                }
-            }, '');
-            const style = node.cloneNode(false);
-            style.textContent = css;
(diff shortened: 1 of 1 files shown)


PaulJThompson:
@cschramm thanks for approving my PR. However, I don't have write access to this repos. Do you?

cschramm:
No, I just wanted to express that I run into the same issue and that the solution seems fine to me.

PR #1699, opened 2018-11-05
ci: test Node.js 4, 6, 8, 10 and 11

DanielRuf:
We should test the supported versions (see `engines` field in package.json).
Also just even releases are supported and LTS, 7 and other odd release numbers already reached their EOL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: node_js
 node_js:
-- '7'
+- 'node'
+- '10'
+- '8'
+- '6'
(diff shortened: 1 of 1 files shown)


DanielRuf:
Does not support Node.js 4: https://travis-ci.org/niklasvh/html2canvas/jobs/450881493

niklasvh:
Sorry for the late reploy, I am moving the project away from travis-ci to azure pipeline (for better browser testing and faster builds), so i'll close this

PR #1730, opened 2018-12-29
background image bug fix

funoy:
fix the bug when config is  useCORS: true and background is cross domain image


niklasvh:
The image is `this.options.allowTaint === true` or `isInlineImage(src)`  or `this.isSameOrigin(src)` so CORS shouldn't be applied.

PR #1744, opened 2019-01-23
check if resize dimensions NaN

ahebrank:
Similar to #1743 , we find that only in Firefox requested image dimensions can be NaN. This is a quick check before drawImage is called.


niklasvh:
How could this be reproduced?

PR #1851, opened 2019-05-26
fix: multi token overflow

### LINKED ISSUE #1850: Element with style 'overflow: hidden auto' will fail rendering

### Bug reports:
Trying to export an element with style `overflow: hidden auto`

**v1.0.0-rc.0:** works as expected [see here](https://jsfiddle.net/ddhauser/2uzm3afs/).
**v1.0.0-rc.1:** will fail with `Error parsing CSS component value, multiple values found when expecting only one` [see here](https://jsfiddle.net/ddhauser/d2y9u73b/).

### Specifications:

 * html2canvas version tested with:
   * Browser & version: Google Chrome Version 74.0.3729.157 (Official Build) (64-bit)
   * Operating system: MacOS Mojave 10.14.5

niklasvh:
fixes #1850


daniel-hauser:
When are you planning to release a new version?

PR #1863, opened 2019-05-30
fix: zero size iframe rendering (#1668)

### LINKED ISSUE #1668: Cannot render iframe content
version: 1.0.0-alpha.12
browser: chrome 69
os: Mac

The process is stuck when I add to the rendered element one of these iframes: 
`<iframe src="images/logo-icon.svg"></iframe>`
`<iframe src="data:text/html,<p>Some HTML</p>"></iframe>`

As far as I see it is stuck at function ResourceLoader.ready, line 187
Some of the promises remains pending forever and line `return Promise.all(values)` never reached.
If I change the original code to this one:
```JS
var values = keys.map(function (str) {
    return Promise.race([_this5.cache[str].catch(function (e) {
        if (process.env.NODE_ENV !== 'production') {
            _this5.logger.log('Unable to load image', e);
        }
        return null;
    }), new Promise(function(resolve, reject) {
        setTimeout(resolve, 5000, null);
    })]);
});
```
the file is rendered but the iframe area remains blank.

garaboncias:
**Summary**

- if an iframe has not specified width or height it parsed to NaN
- Canvas drawImage throw error when try to draw 0 size image

Fixes #1668

### DIFF
--- a/src/dom/replaced-elements/iframe-element-container.ts
+++ b/src/dom/replaced-elements/iframe-element-container.ts
@@ -15,8 +15,8 @@ export class IFrameElementContainer extends ElementContainer {
     constructor(iframe: HTMLIFrameElement) {
         super(iframe);
         this.src = iframe.src;
-        this.width = parseInt(iframe.width, 10);
-        this.height = parseInt(iframe.height, 10);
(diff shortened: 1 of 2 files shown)


niklasvh:
thanks!

PR #1968, opened 2019-08-13
supports video capturing & fixes bugs

yanfangyao:
**Summary**

* [ ] Feature supports video capturing
                 - Implementation: 
                   Step 1: In cloning period, using canvas api drawImage to draw the current video frame to a temporary canvas. 
                   Step 2: Translating canvas to an inline image.  
                   Step 3: In cloning period, replacing the origin video tag with the img tag generated in Step 2.
                   Step 4: Applying css styles to img tag according to video tag's computed styles. Styles applied to img tag should have the '!important' suffix to prevent user defined styles overwriting them.
                   Step 5: In canvas rendering period or foreignObjectRendering period, the render can recognize the img tag and then rendering it to the final canvas.

* [ ] Bug 1 displaying inline-block & inline-table not correctly.
                  - see test case /tests/reftests/inline-block-inline-table.html

* [ ] Bug 2 fix function parseStatckTree
                  - elements in negativeZIndex or positiveZIndex should be in ascending order according to their zIndex value

* [ ] Bug 3 fix canvas globalAlpha
                  - When calling applyEffects, canvas may be restored to an older state. As a result, canvas's globalAlpha value set in renderStack function may be overwritten. This causes visual problems in some cases.


yanfangyao:
Sorry, I can not figure out what's wrong with video capturing test cases in a short time. They always fail on certain platform. However, my application only needs to work in Chrome, i will close the pr and find out the reason latter.

grassy:
Hello, my application need to capture video, and only needs to work in Chrome, can you send me the builds. Thank you very much.

PR #2011, opened 2019-09-22
fix: nested z-index ordering

niklasvh:
Fixes #1978


rupid:
The problem still exists

niklasvh:
@rupid Could you please share an example where this still exists?

PR #2020, opened 2019-09-27
feat: support Web Components

### LINKED ISSUE #865: Support for shadow root
@niklasvh thanks for this awesome library it really helped to achieve a demanded feature for an open source project here https://github.com/puranjayjain/chrome-color-picker or https://atom.io/packages/chrome-color-picker.

I just wanted to ask if there was by any change support for the ::shadow or shadow root elements as the project uses custom elements which need to be present in the list. Any help would be greatly appreciated.

mcolemanNOW:
Added support for custom elements, shadow DOM, and slots

Fixes #865

### DIFF
--- a/src/dom/document-cloner.ts
+++ b/src/dom/document-cloner.ts
@@ -20,6 +20,7 @@ import {CounterState, createCounterText} from '../css/types/functions/counter';
 import {LIST_STYLE_TYPE, listStyleType} from '../css/property-descriptors/list-style-type';
 import {CSSParsedCounterDeclaration, CSSParsedPseudoDeclaration} from '../css/index';
 import {getQuote} from '../css/property-descriptors/quotes';
+import {nodeListToArr} from '../core/util';
 
 export interface CloneOptions {
     id: string;
@@ -294,14 +295,14 @@ export class DocumentCloner {
             const counters = this.counters.parse(new CSSParsedCounterDeclaration(style));
             const before = this.resolvePseudoContent(node, clone, styleBefore, PseudoElementType.BEFORE);
 
-            let children = node.shadowRoot ? Array.from(node.shadowRoot.childNodes) : Array.from(node.childNodes);
+            let children = node.shadowRoot ? nodeListToArr(node.shadowRoot.childNodes) : nodeListToArr(node.childNodes);
             for (let i = 0; i < children.length; i++) {
                 let child = children[i] as Node;
                 if (isSlotElement(child)) {
                     let assignedNodes = child.assignedNodes() as ChildNode[];
                     if (assignedNodes.length > 0) {
                         // replace the slot element with its assigned nodes, and then process them immediately
-                        children.splice(i, 1, ...Array.from(assignedNodes));
+                        children.splice(i, 1, ...assignedNodes);
                         i -= 1;
                         continue;
                     }
--- a/src/dom/document-cloner.ts
+++ b/src/dom/document-cloner.ts
@@ -9,6 +9,8 @@ import {
     isSelectElement,
     isStyleElement,
     isTextareaElement,
+    isSlotElement,
+    isCustomElement,
     isTextNode
 } from './node-parser';
 import {Logger} from '../core/logger';
@@ -127,7 +129,26 @@ export class DocumentCloner {
             return this.createStyleClone(node);
         }
 
+        if (isCustomElement(node)) {
+            return this.createCustomElementClone(node);
+        }
+
         return node.cloneNode(false) as HTMLElement;
+    }
+
+    // custom elements are cloned as divs so the renderer knows how to render them
+    createCustomElementClone(node: HTMLElement): HTMLElement {
+        let customDiv = document.createElement('div');
+        if (typeof node.getAttributeNames === 'function') {
+            let attrNames = node.getAttributeNames();
+            for (let i = 0; i < attrNames.length; i += 1) {
+                let attrName = attrNames[i];
+                let attrValue = node.getAttribute(attrName) as string;
+                customDiv.setAttribute(attrName, attrValue);
+            }
+        }
+
+        return customDiv as HTMLElement;
     }
 
     createStyleClone(node: HTMLStyleElement): HTMLStyleElement {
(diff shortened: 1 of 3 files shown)


niklasvh:
Thanks for the PR. Could you add some tests to `tests/reftests` so that the output can be verified on different browsers?

aaronclong:
@niklasvh and @mcolemanNOW I was actually working on something similar as well. It looks like these changes don't preserver the shadow dom's styles.

mcolemanNOW:
@aaronclong do you have an example where the styles aren't preserved? From my testing it looked like the styles in the shadow tree were being applied correctly

aaronclong:
@mcolemanNOW I can't share the screenshots because it's from a proprietary web app at my company.

However, in your diff I don't see you copying over the `CssStyleSheet` object in the new shadowRoot under the `adoptedStylesSheet` field. This is where I was running into problems with shadow root scoped styles. You can't share this object between doms because of security concerns. You have to instantiate it with the iframe's dom and assign it to the shadow root of the new element.

aaronclong:
How this library waits till the last minute to create the iframe made this difficult for me.

mcolemanNOW:
It looks like `adoptedStyleSheets` is an experimental feature and is only supported in Chrome 73 and Opera 60

https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot#Browser_compatibility

Maybe we want to address that in a different PR once it is more widely supported?

aaronclong:
@mcolemanNOW that is fair

UniqueAlex:
Failed to set the 'adoptedStyleSheets' property on 'ShadowRoot': Sharing constructed stylesheets in multiple documents is not allowed

I seem to have this problem, may I know how to fix it?

wysj:
how this pr publish？

nehachaudhary:
@niklasvh Could you merge this PR and release the version, kind of stuck in my project because there is no support for the custom elements?

ibrahimAboelsuod:
When is this planned to be merged? I didn't understand what is the delay.

skLeo:
@niklasvh When the feature can be released. My new project is based on Polymer. Thanks

de-don:
Any news?

shripatki:
hello we need this 
please merge this PR

atulhm:
I am using ionic and also am unable to use html2canvas for this reason.  Its a blocking issue for a feature, if I can help in some way please let me know.

shripatki:
@niklasvh please please merge this PR we need this

bitflower:
Is there anything we can help with @niklasvh to get this out?

bitflower:
I thought about this change again. Maybe it would be good to have it configurable? Some scenarios might require to actually have the components reloaded.

src/dom/document-cloner.ts
```diff
@@ -127,9 +130,28 @@ export class DocumentCloner {
...
+        if (isCustomElement(node)) {
+            return this.createCustomElementClone(node);
+        }
+
         return node.cloneNode(false) as HTMLElement;
     }
 
+    // custom elements are cloned as divs so the renderer knows how to render them
+    createCustomElementClone(node: HTMLElement): HTMLElement {
+        let customDiv = document.createElement('div');
```
pmoleri, line 142: Something to consider is that `<div>` defaults to display `block` whereas custom-elements default to display: `inline`.
In some occasions this is not a problem, for example a child of a flex will always be considered a block, so you won't notice any difference.

Is it really necessary to convert custom-elements to a standard element?
I'm using Angular and the original library `1.0.0-rc.5`  didn't have any problem with custom-elements, the only problem was with the ones that use shadow dom encapsulation. When I turn to this fork, it supports shadow dom, although it did break height for one of the elements because of this inline vs block inconsistency. It was easy to fix with an onclone hook though.

In any case, switching to this fork was definitely an improvement for my use case, thanks for this contribution.
I should clarify that I'm not using _slots_ so I can't give feedback on that part.

bennypowers:
> It looks like `adoptedStyleSheets` is an experimental feature and is only supported in Chrome 73 and Opera 60
> 
> https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot#Browser_compatibility
> 
> Maybe we want to address that in a different PR once it is more widely supported?

LitElement (one of if not the most popular web component authoring libraries) uses `adoptedStyleSheets` on supported browsers, which I believe at this point includes Chrome, Edge, and Safari.

If support delays this PR it should be deferred, however adding support will serve a large variety of use cases

alexandreLavenant:
Any news ? it would be greatly appreciated

pmoleri:
@niklasvh @mcolemanNOW 
Is there intention of seeing this merged?

My comments on this:

**custom-elements** works fine for me, the only problem I see is converting custom elements to `<div>` it doesn't match the spec, because custom elements are inline by default while `div` is block by default. I think it would be better to use span or not converting them at all. It actually works good for me if I don't convert them (I'm using Angular).
**shadow dom**  it also works fine for me. Again, I'm using Angular and I think it automatically copies styles to the shadow dom to adopt them.
**slots** I'm not using them so I haven't tested.
**adoptedStylesheets** still not standard.

My recommendations:
   - clone custom-elements to `<span>` instead of `<div>`
   - optional: make _createCustomElementClone_ optional using an options flag
   - merge this PR (it's useful to a lot of people)
   - request a new PR to use adoptedStylesheets under an options flag

conceptree:
Hi @niklasvh & @mcolemanNOW, hope that you guys are doing well. 

Reaching out to ask if this merge will be done soon? 

Cheers.

diegorochawebdg:
Any news?

niklasvh:
Thanks @pmoleri for investigating this. 

As I mentioned in https://github.com/niklasvh/html2canvas/pull/2020#issuecomment-536393349, before any support for web components can/will be added to this repository, there needs to be some reftests in place for them so that any functionality can be tested and make sure it works cross-browser, as well as doesn't end up getting regressions over the years. 

I'd be happy to accept a PR that just adds reftests for web components without support for the actual functionality, which can be followed up by a different PR (this or another approach).

PR #2049, opened 2019-10-23
fixed aws s3 with html2canvas - CORS issue with multiple browsers

pavlovandreyka:
fixed aws s3 with html2canvas - CORS issue with multiple browsers

### DIFF
--- a/src/core/cache-storage.ts
+++ b/src/core/cache-storage.ts
@@ -127,7 +127,13 @@ export class Cache {
             //ios safari 10.3 taints canvas with data urls unless crossOrigin is set to anonymous
             if (isInlineBase64Image(src) || useCORS) {
                 img.crossOrigin = 'anonymous';
+                img.setAttribute('crossOrigin', 'anonymous');
+                src  = src +'?v='+new Date().getTime();
+                img.src = src;
             }
+            img.crossOrigin = 'anonymous';
+            img.setAttribute('crossOrigin', 'anonymous');
+            src  = src +'?v='+new Date().getTime();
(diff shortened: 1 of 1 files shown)


yoyoys:
it should be merge

w-ryba:
Issue still present. Would be nice to merge this.

andrushkodmytro:
Please merge this changes

TimXiaoo (approved):

kimsehwan96:
If you're still experiencing this issue. This can be your solution.

https://velog.io/@kimsehwan96/S3-CORS-%ED%97%A4%EB%8D%94-%EA%B4%80%EB%A0%A8-%EC%9D%B4%EC%8A%88-%ED%95%B4%EA%B2%B0%EB%B0%A9%EB%B2%95-html2canvas-lottie

If you can attach your s3 bucket to cloudfront, you can transfer `origin` header with cloudfront.

스크린샷 2021-05-20 오전 12 06 55

Add custom origin header in cloudfront. 

`Origin` can't be added but 'origin' can.

Add 'origin' : '*' in your cloudfront custom header.

Then you can receive `access-control-allow-origin` header from `s3. bucket`

SteveZhangF (approved):

niklasvh:
The fixed proposed here basically invalidates all image caches, regardless of if you use AWS or something else. From the ticket and the fix proposed, I am having trouble understanding what the issue is, which would help to figure out whether there would be a better solution for this, because as-is, I will not be merging this due to the reason I mentioned.

Issue #1868, opened 2019-06-01
Clone node: Setting className for SVG element raises error


KagamiChan:
Please make sure you are testing with the latest [release of html2canvas](https://github.com/niklasvh/html2canvas/releases). 
Old versions are not supported and issues reported for them will be closed.

# Please follow the general troubleshooting steps first:

- [x] You are using the latest [version](https://github.com/niklasvh/html2canvas/releases)
- [x] You are testing using the non-minified version of html2canvas and checked any potential issues reported in the console

### Bug reports:

Setting className for SVG element will raise error, here's some stack trace, I'm using the non minified js version. I thinks the reason is probably I'm using fontawesome 5 SVG icon

```
MY_PROJECT_PATH/lib/html2canvas.js:5304 Uncaught (in promise) TypeError: Cannot assign to read only property 'className' of object '#<SVGSVGElement>'
    at DocumentCloner.resolvePseudoContent (VM4414 html2canvas.js:5304)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5200)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at DocumentCloner.cloneNode (VM4414 html2canvas.js:5207)
    at new DocumentCloner (VM4414 html2canvas.js:5006)
    at VM4414 html2canvas.js:6900
    at step (VM4414 html2canvas.js:83)
    at Object.next (VM4414 html2canvas.js:64)
[...]

PR #2034, opened 2019-10-15
fix: #1868 Clone node, Setting className for SVG element raises error

hal1984:
**Summary**

This PR fixes/implements the following **bugs/features**

* [ ] #1868

Explain the **motivation** for making this change. What existing problem does the pull request solve?

See #1868. The definition of className differs for Element and SVGElement. In Element className it is a string, but in SVG element class name it is a instance of SVGAnimatedString.

References: [https://developer.mozilla.org/en-US/docs/Web/API/Element/className](https://developer.mozilla.org/en-US/docs/Web/API/Element/className)
[https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString)

**Code formatting**

Please make sure that code adheres to the project code formatting. Running `npm run format` will automatically format your code correctly.

**Closing issues**

fixes #1868
closes #1868

PR #2079, opened 2019-11-26
fix: #1868 Clone node, Setting className for SVG element raises error

niklasvh:
Small update to #2034


Turbotailz:
I would appreciate if the NPM package could be published with this merge. I've had to publish to my own NPM account just to avoid the bug this fixes. Cheers.

Issue #2088, opened 2019-12-12
text-decoration issue


Bitzo:
### CSS `text-decoration: underline` issue

here is jsfiddle demo: http://jsfiddle.net/4f13ed8u/3/

same code but different result on different iOS version 

here are test cases,  iOS 13 shows right output

html2canvas version: 1.0.0-rc.5

- iOS 11.4
![image](https://user-images.githubusercontent.com/18672871/70713716-89b9c180-1d21-11ea-8098-39ff49dee307.png)

- iOS 12.1
![image](https://user-images.githubusercontent.com/18672871/70713738-9807dd80-1d21-11ea-808d-93d9f0862630.png)

- iOS 13.3
![image](https://user-images.githubusercontent.com/18672871/70713988-2f6d3080-1d22-11ea-8d1a-c42d8ae78cda.png)

PR #2090, opened 2019-12-13
fix: text-decoration issue

Bitzo:
this pull request is to  solve issue #2088 

from my debug, on iOS 11.4 and 12.1,   style `text-decoration` value in `declaration.textDecorationLine` is `undefined` or `none`, but can get with `declaration.textDecoration`


niklasvh:
Addressed in https://github.com/niklasvh/html2canvas/pull/2567

PR #2092, opened 2019-12-17
Remove line about library being experimental

andrewbloom37:
Seeing the number of users of this library and length of time it has been around, I don't think it is necessary to describe it as "experimental" any more.


niklasvh:
Just because users choose to ignore the experimental nature of this library, doesn't mean that it isn't. I still wouldn't recommend using it in production

PR #2100, opened 2019-12-27
Allow proxy url with parameters

### LINKED ISSUE #2098: Proxy url cannot have parameters
I am using a proxy url that contains a parameter:
http://abc.com/myproxy.jsp?myparam=foo

When I pass this proxy as an option to html2canvas, my proxy gets the following parameter:
myparam=foo?url=abc.com....
The reason is that the code automatically puts a question mark (?) after the proxy url. The code should check whether there is already a parameter at the end of the url. This was done in the the 0.5 version:
`proxyUrl + (proxyUrl.indexOf("?") > -1 ? "&" : "?") + "url=" + encodeURIComponent(src)`

Please fix.

mzand33:
**Summary**
Currently, proxy url's cannot have parameters. 

2098

If a proxy url contains a parameter such as:
http://abc.com/myproxy.jsp?myparam=foo
html2canvas will call the proxy using the following url:
http://abc.com/myproxy.jsp?myparam=foo?url=abc.com....
Notice there are two question marks in the url which is incorrect. The second parameter should begin with an ampersand.
The code had automatically put a question mark (?) after the proxy url. I fixed the code to check whether there is already a parameter at the end of the url. This was properly done in the the 0.5 version:

**Test plan (required)**
Pass a proxy url which ends with a parameter(s) as an option.

**Code formatting**

**Closing issues**

Fixes #2098
Closes #2098

### DIFF
--- a/src/core/cache-storage.ts
+++ b/src/core/cache-storage.ts
@@ -179,7 +179,7 @@ export class Cache {
 
             xhr.onerror = reject;
             const query_string = proxy.indexOf('?') > -1 ? '&' : '?';
-			xhr.open('GET', `${proxy}${query_string}url=${encodeURIComponent(src)}&responseType=${responseType}`);
+            xhr.open('GET', `${proxy}${query_string}url=${encodeURIComponent(src)}&responseType=${responseType}`);
 
             if (responseType !== 'text' && xhr instanceof XMLHttpRequest) {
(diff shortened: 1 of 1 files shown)


Akshay-Bindal:
What is the status of this PL? Can we please expedite this as it seems like a common use case where proxy url can have existing query params.

PR #2109, opened 2020-01-12
Position text relative to baseline

ljsnagy:
**Summary**

Currently text position is calculated by using the absolute bounding height and placing it at the bottom. This causes fonts to be misaligned, as well as underlines being misplaced with some fonts too. Using the baseline is far more consistent across fonts.

See <a href="https://jsfiddle.net/Lxn6b7r3/">this fiddle</a>


calnholt:
@niklasvh Would love to see this merged

PR #2395, opened 2020-10-21
Add optional withCredential ResourceOption for CORS images

delewis13:
**Summary**

Problem:
- when requesting a cross-origin image that required cookie based authentication, image retrieval fails as cookies are not attached to GET request

Solution:
- add optional 'withCredentials' toggle to ResourceOptions. Sets `image.crossOrigin` to `use-credentials` when true, otherwise leaves as `anonymous`.

**Test plan (required)**

No tests planned. Would require standing up adding a cookie-auth based mock server for images.


delewis13:
@niklasvh possible to merge this?

PR #2409, opened 2020-11-03
Ensure resizeImage's canvas has at least 1px of width and height

### LINKED ISSUE #2408: Error when drawing images with fractional heights that scale to less than 1px
### Bug reports:

When drawing a background image with an intrinsic width of `1300` and height of `1`, scaled down to a width of `650` (using background-size), the height of the image becomes `0.5`, which once assigned to canvas.height, gets rounded to `0`

https://github.com/niklasvh/html2canvas/blob/3982df1492bdc40a8e5fa16877cc0291883c8e1a/src/render/canvas/canvas-renderer.ts#L535-L537

This then triggers an error of

```
Failed to execute 'createPattern' on 'CanvasRenderingContext2D': The image argument is a canvas element with a width or height of 0.
```

## possible solution  

Switch to use `canvas.height = Math.max(1, height); ` in `CanvasRenderer.resizeImage`

### Specifications:

 * html2canvas version tested with: 1.0.0-rc.7
 * Browser & version: Chrome Version 86.0.4240.111 (Official Build) (x86_64)
 * Operating system: MacOS

cheapsteak:
**Summary**

* [x] fixes Bug #2408

`CanvasRenderer.resizeImage` throws when passed in an image that, after resizing, has a height of less than 1px

**Test plan (required)**

Added a test for background image slivers to reftests/background/size.html

Have tested that test fails with `Failed to execute 'createPattern' on 'CanvasRenderingContext2D'` error before the fix, and passes after the fix

**Closing issues**

Fixes #2408


abdurrahmanyildirim:
Hi @cheapsteak , Could you update your solution with this ? 
``` 
canvas.width = isNaN(width) || width === 0 ? 1 : width;
canvas.height = isNaN(height) || height === 0 ? 1 : height;
```
At my app, width comes with NaN value and ``` Math.max(1,width) ``` gives NaN result so it gives error.

PR #2529, opened 2021-05-11
feat: added direction css property

### LINKED ISSUE #2488: RTL direction problem
Hi, there is a problem with Arabic/Farsi characters and RTL direction,
the correct HTML file picture:
image

html2canvas render result:
image

hasanparasteh:
**Summary**

* [ ] RTL text formating. It fixes #2488 issue.

When I wanted to use jsPdf, html2pdf.js and etc, there was a problem with RTL text direction in rendering process. So this PR should add support to direction property for `CanvasRenderingContext2D.direction` and fix the problem.

I also created a reftests to demonstrate the problem and showing the issue is existing.


niklasvh:
This PR parses the `direction` property from CSS but never applies it to the `CanvasRenderingContext2D.direction` property?

hasanparasteh:
I just added direction property... now when you said it I realized that I forgot to apply it to `CanvasRenderingContext2D.direction`😅.

Issue #1888, opened 2019-06-13
textarea align issue


xuyangbill:
text align top in the textarea, but align center in the screenshot

### Bug reports:

https://jsfiddle.net/xyjko4b2/

### Specifications:

 * html2canvas version tested with:v1.0.0-rc.3
 * Browser & version:chrome  74.0.3729.157
 * Operating system:win10
![screenshot](https://user-images.githubusercontent.com/5809774/59403933-4fee1980-8dd7-11e9-8611-820ff387827a.png)

662:
+1

untitledsoft:
Hello, any chance to fix this?

nidi3:
Might be a duplicate of #2008

PR #2708, opened 2021-09-20
fix issues with conflicting styles causing baseline for fonts to be wrong.

mikey0000:
**Summary**

When css targeting body, spans and images exist it can cause the baseline to be calculated incorrectly.

for example display: flex on body will cause it to miscalculate the baseline.

* [x] Text alignment issues with miscalculated baseline

When css properties on body change the offset for the container for calculating the baseline, the baseline is wrong and results in text that is not aligned correctly.

**Test plan (required)**

Set body css to flex, can also target span and img but its not necessary. I've added this in the reftests in text and its kind of reflected in my change.

This PR might fix the text alignment issue #1888 but I haven't checked that specific issue.

closing remarks are the styles could use !important but since they are inline I doubt that matters.

Before:
Screenshot 2021-09-20 200145

After:
Screenshot 2021-09-20 200200

### DIFF
--- a/src/render/font-metrics.ts
+++ b/src/render/font-metrics.ts
@@ -27,21 +27,28 @@ export class FontMetrics {
         container.style.fontSize = fontSize;
         container.style.margin = '0';
         container.style.padding = '0';
+        container.style.position = 'absolute';
+        container.style.top = '-1000px';
 
         body.appendChild(container);
 
         img.src = SMALL_IMAGE;
         img.width = 1;
         img.height = 1;
+        img.style.width = '1px';
+        img.style.height = '1px';
 
         img.style.margin = '0';
         img.style.padding = '0';
         img.style.verticalAlign = 'baseline';
+        img.style.display = 'inline-block';
 
         span.style.fontFamily = fontFamily;
         span.style.fontSize = fontSize;
         span.style.margin = '0';
         span.style.padding = '0';
+        span.style.lineHeight = 'normal';
+        span.style.height = 'auto';
 
         span.appendChild(this._document.createTextNode(SAMPLE_TEXT));
         container.appendChild(span);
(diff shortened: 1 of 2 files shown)


niklasvh:
Do you have an example repro for this issue?

kewp:
This worked for me! Thank you so much

mikey0000:
> Do you have an example repro for this issue?

Sorry no but I can try and gather one together, all it should require is setting the styles I changed on a page to something else.

arnaudambro:
Hi! 👋 
      
Thanks @mikey0000 for this PR.

As it's not merged and it seems it's not going to be, today I used [patch-package](https://github.com/ds300/patch-package) to patch `html2canvas@1.4.1` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/html2canvas/dist/html2canvas.js b/node_modules/html2canvas/dist/html2canvas.js
index 84cb30d..8f2c75f 100644
--- a/node_modules/html2canvas/dist/html2canvas.js
+++ b/node_modules/html2canvas/dist/html2canvas.js
@@ -6572,6 +6572,8 @@
             container.style.margin = '0';
             container.style.padding = '0';
             container.style.whiteSpace = 'nowrap';
+            container.style.position = 'absolute';
+            container.style.top = '-1000px';
             body.appendChild(container);
             img.src = SMALL_IMAGE;
             img.width = 1;
@@ -6579,10 +6581,13 @@
             img.style.margin = '0';
             img.style.padding = '0';
             img.style.verticalAlign = 'baseline';
+            img.style.display = 'inline-block';
             span.style.fontFamily = fontFamily;
             span.style.fontSize = fontSize;
             span.style.margin = '0';
             span.style.padding = '0';
+            span.style.lineHeight = 'normal';
+            span.style.height = 'auto';
             span.appendChild(this._document.createTextNode(SAMPLE_TEXT));
             container.appendChild(span);
             container.appendChild(img);
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #2720, opened 2021-10-01
Improve text performance by combining words in the same line

KurtGokhan:
**Summary**

Normally, html2canvas splits the text into words before rendering for word wrapping to work correctly. However this causes too many calls to `ctx.fillText` as each word is rendered separately.

This PR combines multiple words in each line so that only one `fillText` call is done for each line.

Note that this does not change anything if `letter-spacing` is not `0`. Because each letter (grapheme) should be rendered separately in that case.

> Explain the **motivation** for making this change. What existing problem does the pull request solve?

The performance cost itself is normally not a big deal. But I made this fix specifically for this issue in `jsPDF`: https://github.com/parallax/jsPDF/issues/3137

`jsPDF` uses a different canvas implementation and each call to `fillText` increases the performance cost. This PR can reduce the number of render calls about 10 times.

**Test plan (required)**

> Demonstrate how the issue/feature can be replicated. For most cases, simply adding an appropriate html/css template into the [reftests](https://github.com/niklasvh/html2canvas/tree/master/tests/reftests) should be sufficient. Please see other tests there for reference.

This should not change any functionality or how the text is rendered. It just increases the text rendering performance. The effect of rendering performance to an image may be unnoticable. But it will be noticable for users of jsPDF.

**Code formatting**

> Please make sure that code adheres to the project code formatting. Running `npm run format` will automatically format your code correctly.

Done.

### DIFF
--- a/src/css/layout/text.ts
+++ b/src/css/layout/text.ts
@@ -22,7 +22,7 @@ export const parseTextBounds = (
     styles: CSSParsedDeclaration,
     node: Text
 ): TextBounds[] => {
-    const textList = breakText(value, styles);
+    const textList = breakText(value, styles, node);
     const textBounds: TextBounds[] = [];
     let offset = 0;
     textList.forEach((text) => {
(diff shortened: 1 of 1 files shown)


src/css/layout/text.ts
```diff
@@ -86,8 +86,34 @@ const getRangeBounds = (context: Context, node: Text, offset: number, length: nu
...
+
+const breakLines = (value: string, styles: CSSParsedDeclaration, node: Text): string[] => {
+    const words = breakWords(value, styles);
+    const lines: string[] = [];
+
+    let offset = 0;
+    let acc = '';
+
+    for (const word of words) {
+        const rects = createRange(node, offset, acc.length + word.length).getClientRects();
```
niklasvh, line 101: This will be computing the text ClientRects twice now

KurtGokhan: Where else is it computed? As I see, it is calculated once again in the lines 30..45 for the combined line, but here it is calculated for each word separately.

niklasvh: https://github.com/niklasvh/html2canvas/blob/master/src/css/layout/text.ts#L31-L40. It either would need to persist the layout information from the words into the lines, or skip splitting the texts if they are on the same line.

This does not appear to handle [word-spacing](https://developer.mozilla.org/en-US/docs/Web/CSS/word-spacing) currently?

KurtGokhan: It handles word-spacing. This function only decides how to split the text. Instead of word by word, it splits when a word breaks into the new line. It uses `getClientRects` so word-spacing calculation is already done.

I will check if I can reuse the rects in the above function. But I would prefer keeping the logic separate.

PR #2723, opened 2021-10-05
prevent image resolution loss

thesved:
Render and copy images at the current scale, instead of downscaling first thus losing resolution, then rescaling them.

**Summary**

Render images without losing resolution.

This PR fixes/implements the following **bugs**

* [ ] Bug 1: images lose resolution when the scale is greater than 1

Example image generated **after** the fix:
image

Original rendering with version 1.3.2:
image

### DIFF
--- a/src/render/canvas/canvas-renderer.ts
+++ b/src/render/canvas/canvas-renderer.ts
@@ -596,16 +596,29 @@ export class CanvasRenderer extends Renderer {
                 }
 
                 if (image) {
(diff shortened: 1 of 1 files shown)


niklasvh:
Do you have an example repro for this issue?

thesved:
Just created one: https://jsfiddle.net/6e170o3n/

Note: my `devicePixelRatio` is 2
image

PR #2581, opened 2021-07-14
feat: support for custom and slot elements

niklasvh:
Fixes #2020 
Fixes #865
Fixes #430

PR #2751, opened 2021-11-12
Fix missing CustomElement children

andrei-kochetygov:
This PR fixes the following **bugs**:

* [Bug with missing CustomElement children](https://github.com/niklasvh/html2canvas/pull/2581#issuecomment-953280213)

The motivation of these changes is that for now CustomElements is parsing incorrectly.

On [line 18](https://github.com/niklasvh/html2canvas/blob/eeda86bd5e81fb4e97675fe9bee3d4d15899997f/src/dom/node-parser.ts#L18) we calling ``firstChild`` property:
```javascript
for (let childNode = node.firstChild, nextNode; childNode; childNode = nextNode) {
```

However on [line 25](https://github.com/niklasvh/html2canvas/blob/eeda86bd5e81fb4e97675fe9bee3d4d15899997f/src/dom/node-parser.ts#L25) we passing ``childNode`` itself, instead of ``childNode.parentNode`` (the parent of the node assigned to slot):
```javascript
childNode.assignedNodes().forEach((childNode) => parseNodeTree(context, childNode, parent, root));
```
This cause missing ``firstChild`` property on [line 18](https://github.com/niklasvh/html2canvas/blob/eeda86bd5e81fb4e97675fe9bee3d4d15899997f/src/dom/node-parser.ts#L18).

Furthermore, even if we pass ``childNode.parentNode``, we are losing ``slot`` default content nodes as there is no place in code where we are parsing slot tree:
```javascript
<slot>
    <div>Some default content</div>
</slot>
```

We must parse [children](https://github.com/niklasvh/html2canvas/blob/eeda86bd5e81fb4e97675fe9bee3d4d15899997f/src/dom/node-parser.ts#L43) along with [shadowRoot](https://github.com/niklasvh/html2canvas/blob/eeda86bd5e81fb4e97675fe9bee3d4d15899997f/src/dom/node-parser.ts#L41) at the same point:
```javascript
if (childNode.shadowRoot) {
    parseNodeTree(context, childNode.shadowRoot, container, root);
}

if (!isTextareaElement(childNode) && !isSVGElement(childNode) && !isSelectElement(childNode)) {
    parseNodeTree(context, childNode, container, root);
}
```

The reason for this is that children are ***already in LightDOM*** and it is not required to get assigned nodes through the ``slot`` element itself.

Also we must parse ``slot`` as default element, so we must remove this condition:
```javascript
if (isSlotElement(childNode) && childNode.assignedNodes) {
```

### DIFF
--- a/src/dom/node-parser.ts
+++ b/src/dom/node-parser.ts
@@ -39,11 +39,7 @@ const parseNodeTree = (context: Context, node: Node, parent: ElementContainer, r
                     parseNodeTree(context, childNode.shadowRoot, container, root);
                 }
 
-                if (
-                    !isTextareaElement(childNode) &&
-                    !isSVGElement(childNode) &&
-                    !isSelectElement(childNode)
-                ) {
+                if (!isTextareaElement(childNode) && !isSVGElement(childNode) && !isSelectElement(childNode)) {
                     parseNodeTree(context, childNode, container, root);
                 }
             }
--- a/src/dom/node-parser.ts
+++ b/src/dom/node-parser.ts
@@ -21,32 +21,30 @@ const parseNodeTree = (context: Context, node: Node, parent: ElementContainer, r
         if (isTextNode(childNode) && childNode.data.trim().length > 0) {
             parent.textNodes.push(new TextContainer(context, childNode, parent.styles));
         } else if (isElementNode(childNode)) {
-            if (isSlotElement(childNode) && childNode.assignedNodes) {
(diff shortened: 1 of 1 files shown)


niklasvh:
Do you have a test case / repro of where this isn't working as expected? There was some recent changes to how web components are handled in https://github.com/niklasvh/html2canvas/commit/74696faf47c07b48b9c9587db0b999da1c08a8be

PR #2773, opened 2021-12-21
Fix fontMetric baseline 

kooriookami:
The 'TEST' is a text, and the other is a svg. The 'TEST' of canvas is a little bit on the down side.
The dom
image
The canvas
image
The campare
image
image


kooriookami:
@niklasvh Hello, could you please look at this PR?

Issue #2770, opened 2021-12-16
在edge下，生成的图片不会显示文字


potatoShield:
Please make sure you are testing with the latest [release of html2canvas](https://github.com/niklasvh/html2canvas/releases). 
Old versions are not supported and issues reported for them will be closed.

# Please follow the general troubleshooting steps first:

- [x] You are using the latest [version](https://github.com/niklasvh/html2canvas/releases)
- [ ] You are testing using the non-minified version of html2canvas and checked any potential issues reported in the console

### Bug reports:

官网案例在edge下，生成的图片不会显示文字

### Specifications:

 * html2canvas version tested with:
 * Browser & version: Edge 44.17763.831.0
 * Operating system: Windows10

Zhouyex:
same question

PR #2797, opened 2022-01-11
fix: getBoundingClientRect x and y properties in older browsers

MoyuScript:
fix: Properties x and y of BoundingRect is undefined in old browser like Chrome/52.0.2743.100

Fixes: #2770
Fixes: #2733


yilihjy:
Same problem
Reference https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#notes

PR #2806, opened 2022-01-19
fix: Fix documentClone.fonts.ready not resolved in ios13.6

MoyuScript:
Fix `documentClone.fonts.ready` not resolved even though `documentClone.fonts.status` is value of `loaded` in IOS 13.6

It looks like some specific browser version will appear this problem.


xs9627:
Confirm this works. Can this be merged?

yilihjy:
@niklasvh This PR is useful, can it be merged?

PR #2815, opened 2022-01-24
Fix chrome issue with the crossOrigin images

alaa-alshamy:
**Summary**

In Chrome browser if the image loaded before without crossorigin it will be cached and used later even if the next usage has crossorigin so it will fail with CORS error.
so add a random query parameter just to prevent the Chrome from using the cached image.

see more info about the Chrome issue in this link: https://stackoverflow.com/a/49503414

* [x] Bug 1
* [ ] Bug 2
* [ ] Feature 1
* [ ] Feature 2
* [ ] Breaking changes

**Test plan (required)**

Demonstrate how the issue/feature can be replicated. For most cases, simply adding an appropriate html/css template into the [reftests](https://github.com/niklasvh/html2canvas/tree/master/tests/reftests) should be sufficient. Please see other tests there for reference.

**Code formatting**

**Closing issues**

Fixes #

### DIFF
--- a/src/core/cache-storage.ts
+++ b/src/core/cache-storage.ts
@@ -96,6 +96,10 @@ export class Cache {
             //ios safari 10.3 taints canvas with data urls unless crossOrigin is set to anonymous
             if (isInlineBase64Image(src) || useCORS) {
                 img.crossOrigin = 'anonymous';
+                // in chrome if the image loaded before without crossorigin it will be cached and used later even if the next usage has crossorigin
+                // so it will fail with CORS error, so add a random query parameter just to prevent the chrome from using the cached image
+                // see more info about the chrome issue in this link: https://stackoverflow.com/a/49503414
+                src = src + (src.indexOf('?') === -1 ? '?' : '&') + 'cacheBusting=1';
             }
             img.src = src;
             if (img.complete === true) {


wsantosve (approved):

src/core/cache-storage.ts
```diff
@@ -96,6 +96,10 @@ export class Cache {
             //ios safari 10.3 taints canvas with data urls unless crossOrigin is set to anonymous
             if (isInlineBase64Image(src) || useCORS) {
                 img.crossOrigin = 'anonymous';
+                // in chrome if the image loaded before without crossorigin it will be cached and used later even if the next usage has crossorigin
+                // so it will fail with CORS error, so add a random query parameter just to prevent the chrome from using the cached image
+                // see more info about the chrome issue in this link: https://stackoverflow.com/a/49503414
+                src = src + (src.indexOf('?') === -1 ? '?' : '&') + 'cacheBusting=1';
```
wsantosve, line 102: It will only work on the first try, if the user execute again the process... It will fail agin. Here is a better approach:
```suggestion
                src = `${src}${src.endsWith('?') ? '&' : '?'}cacheBusting=${Math.random()}`;
```

wsantosve (changes requested):

aprilandjan:
I found this [good article](https://www.hacksoft.io/blog/handle-images-cors-error-in-chrome) and came here. This is what we need, and can be very useful in most cases 👍

src/core/cache-storage.ts
aprilandjan: `endsWith('?')` should be `includes('?')` 🤔

PR #2832, opened 2022-02-14
Fix foreign object renderer offset

hansottowirtz:
**Summary**

`ForeignObjectRenderer.render` incorrectly calls `createForeignObjectSVG` with `scale` as the parameters for x and y offset. Because of this, it behaves differently from `CanvasRenderer`.

**Test plan**

Render an object with `foreignObjectRendering: true`. It will have an offset of a few pixels.

With `foreignObjectRendering: true`:

Screenshot 2022-02-14 at 18 53 33

With `foreignObjectRendering: false`:

Screenshot 2022-02-14 at 18 54 11

### DIFF
--- a/src/render/canvas/foreignobject-renderer.ts
+++ b/src/render/canvas/foreignobject-renderer.ts
@@ -30,8 +30,8 @@ export class ForeignObjectRenderer extends Renderer {
         const svg = createForeignObjectSVG(
             this.options.width * this.options.scale,
             this.options.height * this.options.scale,
-            this.options.scale,
-            this.options.scale,
+            0,
+            0,
             element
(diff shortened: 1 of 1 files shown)


mjewell:
I noticed the same thing issue, but shouldn't the fix be `this.options.x, this.options.y`

astrid-kg:
Hi!

In the `ForeignObjectRenderer` constructor the origin of the canvas is moved by negative x and y:
`this.ctx.translate(-options.x, -options.y);`

Since x and y is already adjusted for, would 0, 0 not be the correct arguments?

vfa-tanna:
+1

PR #2846, opened 2022-02-25
Comply with Trusted Types

### LINKED ISSUE #2858: Support Trusted Types and escape HTML special characters
### Bug reports:

[html2canvas](https://github.com/niklasvh/html2canvas) has [Trusted Types](https://web.dev/trusted-types/) violation due to use of `document.write` API which is considered a dangerous sink for XSS. This change fixes that and make `html2canvas` compatible with Trusted Types.

The usage of `document.write` in `html2canvas` is also dangerous in a way that, it might have potential to trigger XSS with [DOM Clobbering](https://portswigger.net/web-security/dom-based/dom-clobbering).

E.g.

```
<body>
<form name="doctype"></form>
<script>
console.log(document.doctype); // This returns the form above
</script>
</body>
```

### Specifications:

 * html2canvas version tested with: 1.4.1
 * Browser & version: Chrome 99
 * Operating system: Windows 11

shhnjk:
**Summary**

[html2canvas](https://github.com/niklasvh/html2canvas) has [Trusted Types](https://web.dev/trusted-types/) violation due to use of `document.write` API which is considered a dangerous sink for XSS. This change fixes that and make html2canvas compatible with Trusted Types.

Fixes: #2858

**Test plan (required)**

Existing test should be sufficient as there is no change in the functionality. There should be a test to check for Trusted Types violation, but existing [test.js](https://github.com/niklasvh/html2canvas/blob/master/tests/test.js) has multiple Trusted Types violation, so I wasn't able to use that infra.


tosmolka:
@shhnjk , what about this fix - https://github.com/niklasvh/html2canvas/commit/ac3e5faafeb9ff359f3e8a0a583cded12fc9df0e - so that we avoid maintaining TT policy? Thanks.

shhnjk:
@tosmolka, have you checked if your solution actually picks up things like `internalSlot` or `publicId` from DocType? If that works, I think your solution is elegant :)

tosmolka:
@shhnjk , from my tests it works well with `name`, `publicId` and `systemId` (https://dom.spec.whatwg.org/#concept-doctype). 

I don't think it works with `internalSubset` in newer browsers but that's expected as this property was deprecated and is no longer supported there.

shhnjk:
@niklasvh could you take a look? Thanks!

PR #2869, opened 2022-04-05
fix: replace deprecated String.prototype.substr()

CommanderRoot:
**Summary**

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

**Test plan (required)**

I tested the return of each statement to make sure it's the same as before


Sharcoux:
Can be tested in `@cantoo/html2canvas`

PR #3018, opened 2023-01-12
WIP fix svg with no width height

### LINKED ISSUE #1897: SVG in img tag not rendering correctly in Firefox

- [x ] You are using the latest [version](https://github.com/niklasvh/html2canvas/releases)
- [x ] You are testing using the non-minified version of html2canvas and checked any potential issues reported in the console

### Bug reports:

I am trying to use html2canvas to generate an image from a page that contains an SVG in the form of a base-64 encoded data URI in an `` tag. This has been working fine in Chrome but not in Firefox. I think this may be because Firefox is returning 0 for both `img.naturalWidth` and `img.naturalHeight` which as used as the `intrinsicWidth` and `intrinsicHeight` values in the `ImageElementContainer` class. The SVG has `width` and `height` attributes set. Could the `intrinsicWidth` and `intrinsicHeight` values be derived from the `<svg>`'s  `width` and `height` instead of the ``'s `naturalWidth` and `naturalHeight`?

I don't have an example I can easily provide but I can create a jsfiddle if needed.

### Specifications:

 * html2canvas version tested with: 1.0.0-rc.3
 * Browser & version: Firefox 67.0.2
 * Operating system: Windows 10

sponce:
### Summary 
Fixed handling of svg images with no width/height attributes.
closes #1897

### Details
Such images were not appearing in resulting canvas when using firefox. The details of the problem are explained in https://bugzilla.mozilla.org/show_bug.cgi?id=700533 and https://webcompat.com/issues/64352 but a rough summary is that firefox won't render svg with no width/height attribute via drawImage.
This fix thus recreates the missing attributes from the viewport one when they are missing

### Testing
A new test was added in images/svg called nowidth.html, copied form external.html but using a modified version of image.svg now called imageViewPortOnly.svg and having no width/height attributes.
Running tests before the fix (keep only the first commit of this MR), you would see the generated screenshot for nowidth being empty. With the fix (second commit of this MR) it's working and identical to the original external case.

### Problems remaining
This is still WIP as I had to add a 3rd commit with a bad hack (an extra sleep) as I was not able to debug the remaining issue of timing in the async code.
Here is a copy of the commit message : 

    This adds a sleep in renderReplacedElement so that drawimage works properly when the fix to svg files is used.
    Without that, it looks like drawImage fires too fast and the fix is not yet done (code is async here)
    So far I did not find out where I messed up with async code. Any help would be welcome !

### DIFF
--- a/src/render/canvas/canvas-renderer.ts
+++ b/src/render/canvas/canvas-renderer.ts
@@ -265,29 +265,64 @@ export class CanvasRenderer extends Renderer {
         });
     }
 
-    renderReplacedElement(
+    async fixSVGImage(
+        container: ReplacedElementContainer,
+        image: HTMLImageElement | HTMLCanvasElement
+    ): Promise<HTMLImageElement | HTMLCanvasElement> {
+        if (image instanceof HTMLImageElement && container.intrinsicWidth == 0 && container.intrinsicHeight == 0) {
+            // in such case we will to use the viewport attribute
+            // and we preprocess the svg image and add the width/height svg attributes to it so that drawImage works on firefox
+            // See https://bugzilla.mozilla.org/show_bug.cgi?id=700533 and https://webcompat.com/issues/64352
+            var img: HTMLImageElement = image;
+            var response = await fetch(img.src);
+            var str = await response.text();
+            var doc = new window.DOMParser().parseFromString(str, 'text/xml');
+            var svgElem = doc.documentElement;
+            if (svgElem) {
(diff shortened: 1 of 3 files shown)


Sharcoux:
I don't understand the need for sleep()

sponce:
See the "problems remaining" section. I also do not get why I needed it. I must say I'm not really good at async code so I must have missed something somewhere and the sleep is just  bad hack to make things work

Issue #3041, opened 2023-03-23
The image rendered using the background image attribute has become blurred, and I have configured high resolution to make it unclear


summer-07:
```html
     <!-- Background-Image -->
    <div id="reference" style="display: inline-block; width: 20px; height: 20px;">
        <div style="background-image: url(https://img.coregem.net/img/user/avatar/random/14.png?imageView2/1/w/200/h/200); opacity: 1; width: 20px; height: 20px; background-size: cover;"></div>
        <!-- <img id="img" src="https://img.coregem.net/img/user/avatar/random/14.png?imageView2/1/w/200/h/200" width="100" height="100"> -->
    </div>

    <script type="text/javascript" src="../dist/html2canvas.min.js"></script>
    <script type="text/javascript">
        await html2canvas(reference, { scale: 4, allowTaint: true, useCORS:true }).then(function (canvas) {
            document.body.appendChild(canvas);
        });
    </script>
```

![image](https://user-images.githubusercontent.com/58579560/227098471-7ff14d42-f862-41d0-8abb-a6d95a46ca75.png)

```html
    <!-- This Img tag -->
    <div id="reference" style="display: inline-block; width: 20px; height: 20px;">
        <!-- <div style="background-image: url(https://img.coregem.net/img/user/avatar/random/14.png?imageView2/1/w/200/h/200); opacity: 1; width: 20px; height: 20px; background-size: cover;"></div> -->
        <img id="img" src="https://img.coregem.net/img/user/avatar/random/14.png?imageView2/1/w/200/h/200" width="20" height="20">
    </div>

    <script type="text/javascript" src="../dist/html2canvas.min.js"></script>
    <script type="text/javascript">
        html2canvas(reference, { scale: 4, allowTaint: true, useCORS:true }).then(function (canvas) {
            document.body.appendChild(canvas);
        });
    </script>
```
![image](https://user-images.githubusercontent.com/58579560/227099098-ba5e036d-d7e6-4b5a-9706-b55fca196049.png)

summer-07:
version: html2canvas 1.0.0-alpha.12

Tingglelaoo:
> version: html2canvas 1.0.0-alpha.12

@007-liang version 1.4.1 still caught this problem

PR #3101, opened 2023-08-06
fix: background images should not be blurry in export

### LINKED ISSUE #1464: background images blurry in export

zhbhun:
**Summary**

Background images should not be blurry in export.

**Reproduction**

- https://stackblitz.com/edit/js-zcrji6?file=index.js,index.html

**Closing issues**

Fixes

- #1464
- #3041

### DIFF
--- a/src/render/canvas/canvas-renderer.ts
+++ b/src/render/canvas/canvas-renderer.ts
@@ -600,11 +600,24 @@ export class CanvasRenderer extends Renderer {
                         image.height,
                         image.width / image.height
                     ]);
-                    const pattern = this.ctx.createPattern(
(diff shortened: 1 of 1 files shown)


Sharcoux:
I find those changes highly suspicious. I recommend a double-check.

linjunc:
Hello, why is this mr not merged?

PR #3128, opened 2023-11-06
feat: added field withCredentials to xhr

Artem5725:
**Summary**
This pr adds field [withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) to the options.
**Usage example**
await html2canvas(document.body, { ... withCredentials: true, ... });
**Motivation**
Field must be set to true in case the proxy needs to receive and check cookies


hexiaokang (approved):

bastiengrignon:
Do you think this PR will be merged one day ?

PR #2938, opened 2022-08-18
Update canvas-renderer.ts to fix Vertical alignment issues with text

haztheo:
- Bug 1 Fixes vertical alignment of text within divs - issue shown here: https://github.com/niklasvh/html2canvas/issues/2937

- Now the fillText usage logically makes sense

- Tested across Chrome, Firefox, Safari

Before:

After:

PR #3151, opened 2024-03-04
Vertical text alignment fix

acurrieclark:
An attempt has previously been made to fix this in #2938, but I don't think it worked as intended.

This fix has been working well for me an I hope it can help others.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,8 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "1.4.0",
+      "name": "html2canvas",
+      "version": "1.4.1",
       "license": "MIT",
       "dependencies": {
         "css-line-break": "^2.1.0",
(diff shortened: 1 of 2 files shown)


yorickshan:
thx for your pr, this fix has been merged to [html2canvas-pro](https://github.com/yorickshan/html2canvas-pro)

matortheeternal:
@yorickshan This change broke font rendering by html2canvas-pro when using a non-default line height.

Specifically: `fontSize` is not the pixel offset of the baseline of text when `lineHeight` is not = to the browser default of 1.175.

I have not verified that the old baseline returned by getFontMetrics returns the correct value, but I believe it is more logically correct than simply assuming fontSize is the distance from the top of the rect to the baseline, and a simple test can show this.  Simply set `line-height: 1` and you will observe a 17.5% difference in the position of text when comparing the browser to the canvas produced by html2canvas-pro.

